PR #27, opened 2015-12-04
Update No-Intro DAT files

RobLoach:
This updates the DAT files with the latest from No-Intro's datomatic.


dat/Atari - 7800.dat
```diff
@@ -1,6 +1,7 @@
 clrmamepro (
 	name "Atari - 7800"
 	description "Atari - 7800"
+	header "No-Intro_A7800.xml"
```
RobLoach: Is there a way that we could support using the `xml` header information?

dat/Atari - Lynx.dat
```diff
@@ -1,6 +1,7 @@
 clrmamepro (
 	name "Atari - Lynx"
 	description "Atari - Lynx"
+	header "No-Intro_LNX.xml"
```
RobLoach: Same goes for Lynx here.

twinaphex:
No XML dependencies please.

RobLoach:
`Nintendo - Famicom Disk System.dat` and `Nintendo - Nintendo Entertainment System.dat` have `header` XML dependencies. They should be removed. Not quite sure how to remove the header skip, while keeping the updates in place.

sergiobenrocha2:
The Air Cars from Jaguar has wrong CRC in the no-intro database, take a look at the PR #21

sergiobenrocha2:
MSX2 is misspelled as MSX 2 in dat/Microsoft - MSX2.dat

You should take a look at the previous commits, you are reverting some fixes here.

sergiobenrocha2:
You reverted 062bb84 too, PC Engine SuperGrafx is misspelled too

RobLoach:
> MSX2 is misspelled as MSX 2 in dat/Microsoft - MSX2.dat, 3a77b0d

> The Air Cars from Jaguar has wrong CRC in the no-intro database, take a look at the PR #21

[Datomatic](http://datomatic.no-intro.org/) distributes as "Microsoft - MSX 2", and "NEC - Super Grafx". Should we rename them in the other files to match what No-Intro pushes out? Or follow our own convention? Thoughts?

sergiobenrocha2:
Their convention is wrong... and some CRC too.

sergiobenrocha2:
You should use only one commit for each file, it's hard to see the diff here

sergiobenrocha2:
Have you tested the NES database? If it's based in the headerless database, it'll not work with Nestopia, only roms with header works there. I fixed some CRC for super mario bros. and can't see it in your PR

sergiobenrocha2:
About the misspelled:

MSX2: http://www.msx.org/wiki/MSX2

PC Engine SuperGrafx: https://en.m.wikipedia.org/wiki/PC_Engine_SuperGrafx

RobLoach:
> You should use only one commit for each file, it's hard to see the diff here

Agreed.

> About the misspelled:
MSX2: http://www.msx.org/wiki/MSX2
PC Engine SuperGrafx: https://en.m.wikipedia.org/wiki/PC_Engine_SuperGrafx

Would be good to fix these in the No-Intro upstream. Could post in [the No-Intro Forums](http://forums.no-intro.org/) about it.

> Have you tested the NES database? If it's based in the headerless database, it'll not work with Nestopia, only roms with header works there. I fixed some CRC for super mario bros. and can't see it in your PR

dat/Atari - ST.dat
```diff
@@ -1579,6 +1579,12 @@ game (
 )
 
 game (
+	name "Stone Age (Europe) (Amiga + ST)"
```
RobLoach: Should probably remove private roms like this too.

twinaphex: We could put them in a separate dat, that way they get combined and you don't need to keep modifying the main DATs like this. I think that would probably be a more sustainable route.

RobLoach: I don't think a separate dat for the private roms would help. Keeping the private roms in doesn't hurt anything.... Having a separate DAT for our customized changes from the upstream, however, would. The [Air Cards (World) change](https://github.com/libretro/libretro-database/pull/27/files#diff-d62c26b1fa607433b57e1c065df51177R41), for instance, would benefit from placing in a different DAT.

dat/Microsoft - MSX2.dat
```diff
@@ -1,7 +1,7 @@
 clrmamepro (
-	name "Microsoft - MSX2"
-	description "Microsoft - MSX2"
-	version 20130225-030822
+	name "Microsoft - MSX 2"
+	description "Microsoft - MSX 2"
```
RobLoach: Here is another change where we want to deviate from what No-Intro provides... Moving the override to a different DAT override or something, would benefit us.

RobLoach:
Created pull requests for these individually. In addition, created https://github.com/libretro/libretro-database/issues/62 as a follow up issue.

PR #45, opened 2015-12-12
Update Microsoft - MSX2

RobLoach:


dat/Microsoft - MSX2.dat
```diff
@@ -1,7 +1,7 @@
 clrmamepro (
 	name "Microsoft - MSX2"
 	description "Microsoft - MSX2"
```
RobLoach: The DAT is named "Microsoft - MSX 2", but we are keeping it as "MSX2".

PR #61, opened 2015-12-12
Update DOOM Developer DAT

RobLoach:
Does this work? Share metadata across all the roms?

### DIFF
--- a/metadat/developer/DOOM.dat
+++ b/metadat/developer/DOOM.dat
@@ -40,7 +40,7 @@ game (
 )
 
 game (
-  developer "id Software|TeamTNT"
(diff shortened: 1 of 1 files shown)


twinaphex:
Please test this beforehand before sending PRs like this. It is pretty dangerous to assume things to just work out like this.

twinaphex:
I guess I need to start writing some tutorial about how to use libretro-super's database generation scripts.

The problem with so much of this is how I am completely swamped in dev work and don't have time for all of this auxilliary stuff like writing readmes, etc.

RobLoach:
:+1:

PR #68, opened 2016-01-08
Add Cave Story.dat

RobLoach:
This is the beginnings of indexing the executables from Cave Story that are playable through the libretro NXEngine core.

These two were obtained from http://www.cavestory.org/download/cave-story.php . Running the files through `wine Doukutsu.exe` shows that they're both version 1.0.0.6, while libretro looks to use nxengine 1.0.0.4. Would appreciate adding [more translations if needed](http://www.cavestory.org/download/cave-story.php)!

### DIFF
--- a/dat/Cave Story.dat
+++ b/dat/Cave Story.dat
@@ -1,0 +1,18 @@
+clrmamepro (
+	name "Cave Story"
+	description "Cave Story"
+	version 20160108
+	comment "libretro | www.libretro.com"
+)
+
+game (
+	name "Cave Story (English) (1.0.0.6)"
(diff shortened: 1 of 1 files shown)


dat/Cave Story.dat
```diff
@@ -0,0 +1,18 @@
+clrmamepro (
+	name "Cave Story"
+	description "Cave Story"
+	version 20160108
+	comment "libretro | www.libretro.com"
+)
+
+game (
+	name "Cave Story (English) (1.0.0.6)"
```
RobLoach: Rather than the language, should we use the region? To match the naming convention of No-Intro?

PR #63, opened 2015-12-28
Update Nintendo DS/DSi and Genesis *.rdb, Mega Man 3 (USA) from NES and others

sergiobenrocha2:

PR #72, opened 2016-01-17
Update the RDBs

RobLoach:
This is the current result of running [libretro-super](https://github.com/libretro/libretro-super)'s [`./libretro-build-database.sh`](https://github.com/libretro/libretro-super/blob/master/libretro-build-database.sh).

@sergiobenrocha2 I see you updated some of these in [your other pull request](https://github.com/libretro/libretro-database/pull/63). Think we should wait off on this one, clean up yours first, or cherry-pick individual commits from that to make merging PRs easier.


sergiobenrocha2:
Uh... could you wait the another PR?

PR #125, opened 2016-05-09
Update Do Ra Mi Fantasy.cht

skx7:
cleaned, removed obsolete cheats doing same effect and glitchy cheats


twinaphex:
You need to rebase this cheat PR it seems.

skx7:
It are PAR cheats, seems to be in the right PAR folder, no?

PR #129, opened 2016-05-31
Update Nintendo - Nintendo 64.dat

OfficialGitHubDeveloper:
- Added entries for all verified .z64 ROMS in GoodN64v3.14.
- Added entries for some non-verified games, not including modified ROMS(hacks, translations, fixes, etc.)
- Added a comment in the header.

- Did not add any entries for Public Domain ROMS, however, it is on my roadmap.

I don't know what the 1 deletion is, because my GitHub Desktop Client is giving me errors; I plan to look into it soon.

### DIFF
--- a/dat/Nintendo - Nintendo 64.dat
+++ b/dat/Nintendo - Nintendo 64.dat
@@ -3,18 +3,21 @@ clrmamepro (
 	description "Nintendo - Nintendo 64"
 	version 20151122-045954
 	comment "no-intro | www.no-intro.org"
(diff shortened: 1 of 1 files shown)


Kivutar:
No intro will provide monolithic dats, not incremental updates. Libretro super have some scripts to merge dats at build time based on a given key matching. This would be a convenient way to store your additions while keeping the no intro updating easy.

I just scare than two years later, somebody not aware of your additions simply override the full dat, and then all your work is lost.

PR #152, opened 2016-06-12
[ScummVM] Remove ScummVM metadat files

RobLoach:
Myst is [currently unsupported by ScummVM](http://wiki.scummvm.org/index.php/Myst). We don't need this stuff in here.


twinaphex:
No, you're wrong. It runs perfectly fine on ScummVM core right now.

Please don't rely on that outdated wiki to determine whether games are ready to be played on ScummVM or not.

RobLoach:
> No, you're wrong. It runs perfectly fine on ScummVM core right now.

Actually, having these breaks ScummVM database compilation. We're using `rom.crc`, and the Myst entries don't have `crc`s, so `libretro-build-database.sh` doesn't compile `ScummVM.rdb`.

> Please don't rely on that outdated wiki to determine whether games are ready to be played on ScummVM or not.

The wiki is not outdated. 
```
scummvm --list-games
```
Does not list Myst as a compatible game.

twinaphex:
I have been running Myst for close to a year or more now on ScummVM. Seems good enough to me. Whatever they list as 'supported' has to fulfill all sorts of requirements which are not really important.

PR #170, opened 2016-06-14
Final Burn Alpha dat based on v0.2.97.38 (without BIOS entries)

lollo78:
Now it's more clean!


twinaphex:
@kivutar I think this is ready to go, right?

RobLoach:
Don't want to list the BIOS? Seems fine. :+1:

lollo78:
Yes, listing bios in MAME and FBA it's useless.
;)

PR #189, opened 2016-06-20
Add Spaceship and Lutris

RobLoach:
- https://github.com/Kivutar/lutro-spaceship
- https://github.com/robloach/love-lutris


twinaphex:
how do you actually scan for these lutro games so they get added to a playlist? Tried it yesterday but without success.

PR #224, opened 2016-07-26
Add Sega - Saturn support

RobLoach:
This DAT will allow RetroArch to browse and play Sega Saturn games. It is built from the Sega Saturn DAT from [redump.org](http://redump.org), using the [libretro-database-sega-saturn](https://github.com/RobLoach/libretro-database-sega-saturn) script.

## Caveat

[Yabause currently doesn't  support loading multiple tracks](http://libretro.com/forums/showthread.php?t=4791), so this DAT only writes the first track. It still works with a large majority of games in Yabause, however. If another DAT comes along (or a core that supports multiple tracks), then we could roll with that. But for now, this is the best we have. Just make Yabause load the first track.

### DIFF
--- a/dat/Sega - Saturn.dat
+++ b/dat/Sega - Saturn.dat
@@ -1,0 +1,9331 @@
+clrmamepro (
+	name "Sega - Saturn"
+	description "Sega - Saturn"
+	version "20160705 01-00-24"
+	comment "libretro-database's Sega - Saturn.dat, built from Redump.org"
+	homepage "https://github.com/RobLoach/libretro-database-sega-saturn"
+)
+
+game (
+	name "Lunar 2 - Eternal Blue (Japan) (Disc 1)"
+	description "Lunar 2 - Eternal Blue (Japan) (Disc 1)"
+	rom ( name "Lunar 2 - Eternal Blue (Japan) (Disc 1) (Track 1).bin" size 653957136 crc 49d70180 md5 07136c4431147a54f12a15291102a3ef sha1 135c69c600dac940cd542a3436cb26a72a0929ec )
+)
+
+game (
+	name "Lunar 2 - Eternal Blue (Japan) (Disc 2)"
+	description "Lunar 2 - Eternal Blue (Japan) (Disc 2)"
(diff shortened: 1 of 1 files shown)


markwkidd:
As far as I can glean, the other standard Saturn sets are Trurip (cue/bin), Darkwater (cue/bin). and TOSEC (ISO).

So TOSEC it is?

RobLoach:
> TOSEC

Looks like the last TOSEC Sega Saturn update was in 2012, and it's missing a good amount from Redump. Looks like they split the ISO into multiple tracks as well. We could merge it into the DAT via https://github.com/RobLoach/libretro-database-sega-saturn though. That wouldn't be a bad idea. Allow either redump/tosec.

markwkidd:
This is starting to remind me of the MAME playlist situation, which I have spent more time thinking about (and writing scripts to support) than Saturn. I have been thinking that it would eventually be simplest if there were a few ClrMamePro DATs available to help RetroArch users get their MAME sets into optimal support for the playlist generator.

Maybe there could eventually be a RetroArch Saturn DAT that munges together the supporting the most possible games without running into BIN/CUE files. When I have more time I will try to get a better idea of the set or sets that standalone Yabuse users are working with. Maybe there are some other clues in that regard.

markwkidd:
Now that this has been merged, how should I describe the situation to users in the Lakka docs.

Does this work? "Lakka's Saturn emulator supports Redump-validated Saturn CUE/BIN sets that incorporate only one BIN. CUE/BIN ROMs with multiple BIN tracks are not currently supported."

RobLoach:
Added TOSEC over at https://github.com/libretro/libretro-database/pull/225 .

> Now that this has been merged, how should I describe the situation in the docs?

Rather than "RetroArch's Saturn emulator", just call it "libretro-yabause", since that's the core that is used.

markwkidd:
I didn't find anything speaking directly to the issue of CD image format in the Yabuse docs, so I have been reading through their forums today. 

With the standalone Yabuse emulator users are recommended to use ISO files or mount a drive image to simulate a Saturn CD in a PC drive since Yabuse supports loading from original Saturn CDs that are physically present.

So I imagine the solution for standalone Yabuse on these multiple-bin games is for ISO or to just mount the cue/bin. I'm adding this comment here as a starting point in case this issue comes up again.

PR #228, opened 2016-08-04
Update ScummVM to single rom sets

RobLoach:
Updated ScummVM.dat to have single rom sets, at the request of @Kivutar. Reduces the size of the .DAT by quite a bit. It combines all of Gruby's DATs, and selects just one rom entry through a script over at:
https://github.com/robloach/libretro-scummvm.dat

You can preview the result over at:
https://github.com/RobLoach/libretro-scummvm.dat/blob/master/ScummVM.dat


Kivutar:
Thank you

PR #251, opened 2016-08-19
Update Sony PlayStation to use Redump/TOSEC/Trurip

RobLoach:
@twinaphex enjoys using serials for PlayStation, @Kivutar requested scanning the .cue files through [Redump](http://redump.org/). I'll let you two decide which is best.

This builds from [Redump](http://redump.org/), [Trurip](http://trurip.org/) and [TOSEC](http://www.tosecdev.org/) via [libretro-dats](https://github.com/robloach/libretro-dats). Each has a documented way to dump games from the disc so that scanning works correctly.
## Caveats
- Loosing a lot of the metadata since it doesn't have CRC data. We could manually update it to have the CRC information, but that sounds like a lot of work, that may quickly become unmanageable.


Kivutar:
Pros of switching to this CUE CRC scanning:

 * We can support 100% of Redump, TOSEC, Trurip , which will bring scanning support for:
   * PCE CD
   * Sega CD
   * Saturn
   * PC-FX
 * It's a step forward in supporting zipped CDs.
 * It works with the same code path that is used for cartridges scanning, which is very simple
 * It's faster than scanning the serial
 * Forces users to find good CD images and verify them
 * Easy to maintain the DATs

Cons of CUE CRC scanning:

* We support only the ROMsets from Redump, TOSEC, Trurip. Are they easy to find?
* We need to keep serial scanning anyway for PSP ISOs (but that will not interfere with CUE scanning) so we will not be able to drop that serial scanning code anyway.

Pros of the Serial Scanning:

 * Works on all formats, even IMG or BIN with missing CUE
 * Is needed anyway for PSP ISOs

Cons of the Serial Scanning:

 * Fails to scan 100% of the CDs, even CDs from the Redump romsets. Implementation is either incomplete, or CD images have garbage serials, or our database lacks some serials.
 * Will never allow for zipped CDs
 * **It's difficult/impossible to find serials databases.** And most of them are incomplete. This makes maintaining the DATs difficult. This is the main reason of why we don't already support PCE CD or PC-FX.
 * For each system, we have to use a magic number then parse the ISO to get the serial, which could not always be at the same offset -> complex code

So I would go for switching to this CUE CRC scanning, **only if those ROMsets are easy to find and get**. I would keep Serial scanning for PSP only. This way we get PCE CD, Sega CD, Saturn and PC-FX working, while keeping PSX and PSP working.

cudencuden:
psx cue sheets are downloadable through redump homepage

orbea:
I think another issue of using the serials would be games that have been released in more than one version like suikoden do not scan to the correct version since the serial has not changed.

For example I have both `Suikoden (USA) (v1.0)` and `Suikoden (USA) (v1.1)`, but they both scan as `v1.1`. On the other hand serials allows scanning undubs or fan translations without any problem.

```
game (
        name "Suikoden (USA) (v1.0)"
        rom (
                serial "SLUS-00292"
                image "Suikoden (USA) (v1.0).cue"
        )
)

game (
        name "Suikoden (USA) (v1.1)"
        rom (
                serial "SLUS-00292"
                image "Suikoden (USA) (v1.1).cue"
        )
)
```

ValiantBlade:
Dragon Warrior VII's redump is just flat out not scanning at all for some reason. I assume it has to do with the garbage serials mentioned within the cons above.

RobLoach:
> Dragon Warrior VII's redump is just flat out not scanning at all for some reason.

This Redump update is not merged, yet. RetroArch still uses the serial scanning, until this PR is discussed a bit more and merged.

To test it out, you could download the DAT, build the rdb with libretro-super, and replace the rdb in your own retroarch. Thoughts?

ValiantBlade:
I'm going to see how well the Redump changes work, and I'll report back. I've actually been rebuilding CUEs for the last few days because I was porting all of my PS1 ROMs over to Retroarch, and a majority of them were CUEless, or worse, in the CCD/SUB/IMG formats (only 3 of those ROMS actually came with CUEs, but at least none of them were corrupted CUEs). I can say though, that Redump ROMs are actually easier to find than No-Intro ROMs (I had to go forum digging for those, but not that deep), and I support the Redump updates, but at the same time, it's going to be a pain to redownload ROMs if they don't work (like I expect ALL of the CCD/SUB/IMG ROMs to do).

RobLoach:
The good news is that you can [download the cue files from Redump.org](http://redump.org/).

ValiantBlade:
So, I tried to use Libretro-Super and your Retroarch-Dats to replace the database in Retroarch, and for some reason, even after replacing the dat in the dat folder within libretro-super/retroarch/media/libretrodb and building the database as per the instructions in libretro-db, the rdb remained unchanged and even when I swapped it out in .config/retroarch/database/rdb, nothing changed. Also, in libretro-super/retroarch/libretro-db, dat_converter is broken lua_converter is not there anymore. Apparently in that repo there's a not saying to revert some files, I don't know much aboyt what's happening there, considering I just started to attempt to contribute to the project today.

RobLoach:
I've gone ahead and compiled it for you. Needed to switch the build method in libretro-build-database.sh...
[Sony - PlayStation.rdb.zip](https://github.com/libretro/libretro-database/files/465720/Sony.-.PlayStation.rdb.zip)

Feel free to test it out!

ValiantBlade:
Tested it as well because I had some spare time. Nothing was scanning. Even my redump games with working serials (Tested with Vagrant Story) aren't even scanning. All I did was replace the rdb in my database folder, which I would guess means that you're right and a conflict must be going on if a game that meets both criteria is not working.

RobLoach:
I believe it's RA's task_database.c that's taking over in the scanning:
https://github.com/libretro/RetroArch/blob/master/tasks/task_database.c#L205-L208

Kivutar:
Yes you need to comment a line to disable serial scanning.

ValiantBlade:
I commented out the case cue section and the function itself, because they should probably be left in the file until this feature merges, and to repress the unused function error that resulted when I just commented out the case. All of my Redump ROMs I tested with (Vagrant Story, Dragon Warrior VII, Tales of Destiny II/Tales of Eternia) are loading perfectly but I haven't done a full scan. Some of the older ROMs that worked when serial scanned (Azure Dreams, Lunar 2 Eternal Blue) now no longer work, which kinda tells me that unless you're looking for Redump ROMs they will be very uncommon to have lying around, since most of my ROM collection was built from when before I knew Redump existed.

Here is the revised file: [task_database.c.zip](https://github.com/libretro/libretro-database/files/466244/task_database.c.zip)

RobLoach:
> Yes you need to comment a line to disable serial scanning.

I'll put together a DAT for PSP. Is that the remaining one that uses Serials?

ValiantBlade:
iirc, yes, and that requires a different modification to task_database.c in order to make sure ISO files work. I'll get onto that because it's a quick 5 min fix.

ValiantBlade:
I marked off the names of each file, also, there's something about Lutro files in the same statement, but nothing about serial scanning, so I assume it has something to do with something else.

[[REMOVED] task_database.c.zip](https://github.com/libretro/libretro-database/files/466252/REMOVED.task_database.c.zip)

[[COMMENTED] task_database.c.zip](https://github.com/libretro/libretro-database/files/466254/COMMENTED.task_database.c.zip)

RobLoach:
@ValiantBlade If you click that "Fork" button at the top right of this page, you'll be able to push up your changes to a branch, and a [Pull Request](https://help.github.com/articles/about-pull-requests/) :wink:

ValiantBlade:
Yeah, I'm new to GitHub. I did take a CodeCademy course on Git a while back though.

PR #254, opened 2016-08-20
Add some ps1 dat

orbea:
Am I doing this right?

### DIFF
--- a/dat/Sony - PlayStation.dat
+++ b/dat/Sony - PlayStation.dat
@@ -163,6 +163,22 @@ game (
 )
 
 game (
+	name "Langrisser IV (Japan)"
+	rom (
+		serial "SLPS-01818"
+		image "Langrisser IV (Japan).cue"
+	)
+)
+
+game (
+	name "Langrisser V (Japan)"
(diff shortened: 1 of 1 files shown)


RobLoach:
:+1:

orbea:
Is there anything more I could do to make this PR more complete? I'm still trying to familiarize myself with how libretro-database works so that I can contribute better.

RobLoach:
Nope, you did great. It's good! We have been considering switching from using the serials to CRC .cue checking, but that's another discussion. This PR is good.

orbea:
This commit is tested and works, however it has exposed another issue.
The names on the Little princess discs will be cut off, I suspect game name is just too long...

orbea:
The above issue with the name being cut off is now solved, but it won't detect the boxart for some reason.

Even though the png and dat both seem right...
```
Little Princess +1 - Maru Oukoku No Ningyou Hime 2 (Japan) (Disc 1).cue
Little Princess +1 - Maru Oukoko No Ningyou Hime 2 (Japan) (Disc 1).png
```

PR #255, opened 2016-08-21
Update databases

sergiobenrocha2:
rdb files built again with dat in the end

PR #256, opened 2016-08-21
Rename Fire Red to FireRed

orbea:
These should be `FireRed`, not `Fire Red`. However when trying to make a new rdb with `libretro-build-database.sh` it still scans for `Fire Red` and the rdb file has references to both naming conventions. What am I missing?

### DIFF
--- a/metadat/developer/Nintendo - Game Boy Advance.dat
+++ b/metadat/developer/Nintendo - Game Boy Advance.dat
@@ -12187,7 +12187,7 @@ game (
 )
 
 game (
-	name "Pocket Monsters - Fire Red (Japan)"
+	name "Pocket Monsters - FireRed (Japan)"
 	developer "Game Freak"
 	rom (
 		crc 3B2056E9
@@ -12195,7 +12195,7 @@ game (
 )
 
 game (
-	name "Pocket Monsters - Fire Red (Japan) (Rev 1)"
+	name "Pocket Monsters - FireRed (Japan) (Rev 1)"
 	developer "Game Freak"
 	rom (
 		crc BB640DF7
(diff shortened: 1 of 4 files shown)


metadat/developer/Nintendo - Game Boy Advance.dat
```diff
@@ -12187,15 +12187,15 @@ game (
 )
 
 game (
-	name "Pocket Monsters - Fire Red (Japan)"
+	name "Pocket Monsters - FireRed (Japan)"
```
RobLoach: We should likely follow whatever No-Intro gives us. If you want to push this change forwards to them, it will make maintenance a lot easier.

orbea: I mad this change because everywhere else in the libretro-database already has `FireRed`. Where can I push the changes to No-intro?

RobLoach: On http://datomatic.no-intro.org , you can submit to the Forums, and the Discussion queue.

RobLoach:
This change would get lost in the midst of https://github.com/libretro/libretro-database/pull/255 . To get these kinds of changes in, we should push them to the upstream. If you create a request, mind linking it here?

orbea:
Thanks, needing to submit it to the forums is a bit of a hassle. I will try to remember to do so in the next few days. This is a really only a nitpick, the main issue that caused me to notice this was fixed in sergiobenrocha2's recent PRs for libretro-database and libretro-super.

RobLoach:
http://datomatic.no-intro.org/?page=forum_newpost&action=new is an easy place to submit fixes.

orbea:
On second thought, there seems to be no need as no-intro already has the correct name and its just the retroarch metadat that is wrong?

http://datomatic.no-intro.org/?page=show_record&s=23&n=1616
http://datomatic.no-intro.org/?page=show_record&s=23&n=1672
http://datomatic.no-intro.org/?page=show_record&s=23&n=1341
http://datomatic.no-intro.org/?page=show_record&s=23&n=2641

RobLoach:
Thanks for the investigation! This looks good :+1:

PR #262, opened 2016-08-24
(NDS) New undub dat

orbea:
~~Do not merge yet.~~

I noticed that several of my nds roms are not scanning because they are undubbed so I added them to the dat file. However after making the changes, cloning libretro-super, replacing the dat file with the new one and using libretro-build-database.sh the resulting rdb file does not have any of these undubbed roms.

So my questions are what am I doing wrong? And would you agree that adding these undubbed roms to the dat file is a good idea?

### DIFF
--- a/dat/Nintendo - Nintendo DS Decrypted.dat
+++ b/dat/Nintendo - Nintendo DS Decrypted.dat
@@ -2631,6 +2631,13 @@ game (
 )
 
 game (
+	name "Avalon Code (USA) (Undub)"
+	description "Avalon Code (USA) (Undub)"
+	serial "YOGE"
+	rom ( name "Avalon Code (USA) (Undub).nds size 114851840 crc a54a6a96 md5 65ec54f6f76a49fa1846d92b4c81b26 sha1 a11f2502efa39566e11827a38c6fcc5853d158b2 )
+)
+
+game (
 	name "Avalon Code (Japan)"
 	description "Avalon Code (Japan)"
 	serial "YOGJ"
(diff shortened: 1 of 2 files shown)


RobLoach:
These will get lost on the No-Intro updates. If you move these into [one of the metadat folders](https://github.com/libretro/libretro-database/tree/master/metadat), it will add the entries, while not get lost in the No-Intro updates.

orbea:
I can do that, but I noticed there aren't really any NDS metadat files yet, should I start them from scratch or is there a better way?

RobLoach:
New one, yeah. Coudl likely live in any metadat folder.

orbea:
So I started a developer metadat with only the Undub games and other related games, however they are not added to the resulting rdb file.

RobLoach:
There were some compilation errors. You can see them in the log when running `./libretro-build-database.sh`. Sent up a PR to fix them for you:
https://github.com/orbea/libretro-database/pull/1

nds-undub

These still should be moved out of the main one, but good job on the first few moves :+1:

orbea:
Thanks! This is now working here, I think this should be safe to merge?

twinaphex:
OK, let's try it then.

RobLoach:
> I don't know even why we purged the bios entries, i keep all the bios in a different folder for example, it's not no-intro fault if you guys are messy. It'll be hard to maintain this repo this way

Dat-O-Matic can build the DAT without the BIOS entries.

> we should avoid mess with the content from no-intro...

Agreed, this was merged pre-maturely. Need to move the rest of the entries into the Undub set.

twinaphex:
RobLoach - I will give you admin powers over this then so you can better coordinate this.

RobLoach:
> I will give you admin powers over this then so you can better coordinate this.

Thanks, with the latest PR that you just merged, we should be fine :+1:

twinaphex:
Made you admin now, I will take a laid-back approach to this repo now since you seem to be capable of making more informed decisions.

PR #264, opened 2016-08-24
(PSX) Fix swapped serials

orbea:
Someone accidentally swapped the `Legend of Legaia` demo and full release serials causing the scanner to confuse the full release with the demo and vica versa.

I tested the change with `libretro-build-database.sh` and it worked as expected.

Reference:
http://playstationmuseum.com/tag/scus-94254/
http://redump.org/disc/12310/


RobLoach:
🤗

PR #267, opened 2016-08-28
Remove Undub Nintendo DS from main DAT

RobLoach:
They exist in the metadat/developer DAT.

@orbea Could you confirm you moved them all over?


orbea:
Oh, I did not realize I should of just moved them entirely. My mistake, thanks for fixing that.

PR #273, opened 2016-09-16
Update DAT to 0.177 and add 7z file extensions

Enverex:
Currently the DAT will only pick up the MAME games if they are in zip format rather than 7z, this will fix that. The core itself handles 7z files just fine.


RobLoach:
Have you tested this? I'm not sure it'll work from the way RetroArch operates.

Enverex:
I mimicked the format used for the other DAT files, e.g.

```
game (
	name "'Nam 1965-1975 (Europe) (En,Fr,De,It)"
	description "'Nam 1965-1975 (Europe) (En,Fr,De,It)"
	rom ( name "'Nam 1965-1975 (Europe) (En,Fr,De,It) (Disk 1).ipf" size 1048872 crc 678E4FF4 md5 9D6D34CB47B07AFBA5AC2F741FA39890 sha1 B1A25FAA18670A3B28E2C593DE492CA32FFCBFD5 )
	rom ( name "'Nam 1965-1975 (Europe) (En,Fr,De,It) (Disk 2).ipf" size 1048872 crc 3AE871F5 md5 0CE5BFF656E14196D4DE3C98EFEB2187 sha1 D59388F780C882C81397141A835511BC7A10FF20 )
)
```

... so unless that one also doesn't work, it should work.

RobLoach:
Would you mind compiling it and testing it out? I've had issues with named sets like this before, along with 7z support in RetroArch. Mind giving it a test? Would love to get this in.

Enverex:
Sure, my main setup doesn't use RetroArch for scanning as I was actually doing this for someone else who does use RA's scanner. I'll try and rejig things my side and do a scan to see if it works.

I'll report back when it's done (and I've actually figured out how to compile then).

RobLoach:
There are some notes about how to compile the rdb files over at:
https://github.com/libretro/RetroArch/tree/master/libretro-db#compiling-the-database

Enverex:
Looks like it only uses the last referenced rom, not all of them. The one I used as a reference to make this also wouldn't work for that same reason (and currently doesn't seem to be used).

Is this a limitation of the RA parser, or am I wrong with the syntax here?

Enverex:
I'll close this for now and work on the ideas mentioned, then open a new pull request once I'm happy everything's working properly.

Enverex:
FYI, I created a new folder in metadat, copied my .dat to there with the same name as the original, added the new folder to the build script, built it (which worked and showed that it loaded in the new dat from the new folder with the right name as the parent), checked the rdb which now does indeed contain both the zip and 7z versions.

Unfortunately it still finds nothing when scanning, so I assume there's more to this.

RobLoach:
Just stick it in one of the existing ones, libretro-super only scans the existing folders.

Enverex:
I added it to the build script so that it would still pull it in. I can see the 7z references in the generated rdb so we know it's being seen, surely?

RobLoach:
Ah, nicely done :+1:

Enverex:
It still didn't work though. Something else must be required for them to be picked up.

Will the scanner actually work on archives themselves, or does it always look inside the archives?

markwkidd:
Is any of this also a factor with the MAME 0.78 .DAT I have created here? (Even though mine is .zip) https://github.com/libretro/libretro-database/issues/248

Enverex:
Probably not. The issue we're having here was initially getting it to see more than one set, that's workable by adding the second set to one of the meta folders, but even then, RA still isn't seeing the 7z archives as viable ROMs for the DB, for currently unknown reasons.

PR #274, opened 2016-09-22
Update No-Intro DATs

RobLoach:
This updates all the DATs, except those that require header information:

- Atari - 7800
- Atari - Lynx
- Nintendo - Nintendo Entertainment System
- Nintendo - Famicon


Kivutar:
I was thinking that they stopped referencing headerless roms since 2016?

RobLoach:
> I was thinking that they stopped referencing headerless roms since 2016?

Unfortunately not quite yet, at least in datomatic.

PR #349, opened 2016-12-10
Update Atari - 2600.dat

RetroUp:
I submitted the previous version to No-Intro and they accepted it. This is a slightly updated version.

The following games are still missing.

Actionauts
Birthday Mania
Robot Commando Raid
Scraper Caper

After those are added I think the dat will be 99% complete. Homebrew, alt name dupes, sprite swaps, double ended carts and x-in-1 carts were intentionally left out.


RobLoach:
Nice work, @RetroUp! Loving these updates.

PR #365, opened 2016-12-27
Add DOS DAT proof of concept

RobLoach:
This is a proof of concept on a DOS DAT from the idea at https://github.com/libretro/dosbox-libretro/issues/44. Only [Commander Keen 4](https://en.wikipedia.org/wiki/Commander_Keen_in_Goodbye,_Galaxy) is adopted from the TDC#12 DAT, I hope to automate the rest of it. The process to automate it will be documented over at https://github.com/libretro/dosbox-libretro/issues/44 .


RobLoach:
The DAT parser is erroring out, hold on.

Issue #403, opened 2017-01-22
adding a "source_list.txt" (or "readme.txt") as authoritative list of source DATs


markwkidd:
The forum move wiped out a table I was maintaining in the Lakka forums which helped answer ongoing questions such as: "why won't my ROMs scan?" "what kind of NES ROMs are supported?" "there is not a No-Intro set for Sega CD, what ROMs am I supposed to use?

The new forums don't support tables, so I'm thinking of how I'll reformat the content. It's somewhat readable in its current format: https://forums.libretro.com/t/megathread-which-roms-work-with-lakka/5566

It does seem like there could be an authoritative list somewhere in this repo that users can refer to. It might be pretty handy, and wouldn't be too hard even to maintain by hand if it's not practical to create it in an automated fashion.

Does anyone else see value in adding a file that compiles this information?

RobLoach:
We could add it as part of the repository here. Github markdown supports tables.

markwkidd:
Alrighty, I'll take a crack at it.

Here is a concept for how this could be a manually-managed file in this repo (with a table): https://github.com/libretro/libretro-database/pull/404

PR #404, opened 2017-01-22
Create sources_list.txt as reference for RetroArch's 'parent' databases

markwkidd:
This is a stripped-down version of the table from the old Lakka forum (a version without formatting is available at https://forums.libretro.com/t/megathread-which-roms-work-with-lakka/5566)

It's primitive but perhaps more nuance can be added later if this proves useful.

First discussed in: https://github.com/libretro/libretro-database/issues/403


RobLoach:
If you rename it to `sources_list.md`, GitHub will automatically style it as a table for you.

markwkidd:
Hm. I could swear it was rendering as a table when I submitted the PR. Maybe too much time editing github wikis is making my head parse MD subconsciously.

@RobLoach I just renamed to sources_list.md

PR #407, opened 2017-01-29
[MAME] Add Versioned and TorrentZipped MAME DATs

markwkidd:
This PR would add MAME DATs for each of the 'frozen' MAME cores. I believe that it would be helpful for users and developers alike to have a centralized set of DATs to refer to even though (or perhaps because of!) the fact that there is not full MAME scanning support at present.


RobLoach:
Likely don't need the subfolders. They could probably just all live in `metadat/mame/*` .

markwkidd:
No problem @RobLoach . I just pushed a commit that moves the DATs out of subfolders

RobLoach:
Great thanks! So happy the MAME loading is getting some attention. There are a couple other things to address on this...

1. Having some cleaner filenames (`MAME 0.139.dat`, `MAME 0.159.dat`, etc) would be nice
2. We'll have to go through https://github.com/libretro/libretro-super/tree/master/dist/info and make sure the MAME DATs reference the right .info files
3. The DATs seem to be in XML format, but RetroArch uses clrmamepro format. Will have to make DATs with one entry so that RetroArch knows which file to launch ([libretro-dats](https://github.com/robloach/libretro-dats) project can do this, made an issue https://github.com/RobLoach/libretro-dats/issues/10 )

markwkidd:
No problem to rename the DATs -- I pushed another commit to match the scheme you suggested. Do you think they're OK to merge now?


I'll going to hop over to your libretro-dats ticket in a moment to talk about implementing scanning for these MAME cores. I also just made a new thread about this in the libretro "Development" forum I'll link you to.

larrykoubiak:
Hopefully i won't come across as a killjoy, but pushing MAME dats such as these won't allow scanning as you know. The only work around we have for Arcade DATs at the moment, which we already used for FBA, is to basically get the CRC32 of the zip files and associate titles with those, instead of the underlying rom files, as Retroarch has no rom auditing capabilities.

markwkidd:
Hi @larrykoubiak I'm trying to create a comprehensive description of the arcade ROM set scanning situation in the forum, with the aim of either:

1. adding a new arcade scanning mode that takes advantage of the ZIP format's built-in manifest with CRCs for files contained in the ZIP, or
2. building understanding of the implications of Non-Merged vs. Split and TorrentZip and how those formatting differences should be taken into account when using the current CRC-based scanning method

If you have time to check it out, I'd appreciate your input: https://forums.libretro.com/t/specification-for-arcade-rom-scanning/8624/

twinaphex:
Let me know when this is fit for merging, RobLoach.

markwkidd:
Hi @larrykoubiak and @RobLoach based on what larrykoubiak has been explaining in the forums thread about arcade scanning, I think it is fair to sum up the idea for now as: take the current crc32-based hash scanner as far as it can go with regard to arcade scanning before discussing a completely alternate approach.

To that end, I'm about 2/3 through a process of rebuilding complete Split and Non-Merged sets for all of the frozen MAME cores. (Days of fun!) Once each of the collections validates with the appropriate XML DAT I'm running them through TorrentZip.

From that point, I'm generating three new DAT files per core in the clrmamepro-style with the checksums of the TorrentZip-ed ZIP files. The result will be four 'metadats' per MAME core, which I believe are what we need for further experimentation:

1. **XML DAT** - the "meta-metadat"
2. **BIOS and Device DAT** (TorrentZipped) - Device checksums added for MAME 2014 core onward
2. **Split ROM set DAT** (TorrentZipped, No BIOS, Device, or Mechanical) - Devices introduced by MAME 0.139, Mechanical by 0.159
3. **Non-Merged ROM set DAT** (TorrentZipped, No BIOS, Device, or Mechanical)

I need another day or two to finish processing these ROM sets and generate DATs but in the meantime I did push a commit to this PR earlier which includes the DATs which are already finished generating. It's not a complete set but it might give a good idea of how this is taking form.

markwkidd:
@larrykoubiak this is now everything up through MAME 2015, merged and split, everything torrentzipped. 

Do you already have any scripts to merge multiple dats into one dat and eliminate duplicate name/checksums? Or is there another step to investigate first?

larrykoubiak:
Hey @markwkidd, the script already exist in a way, it's this one : [libretro-build-database.sh](https://github.com/libretro/libretro-super/blob/master/libretro-build-database.sh) that you need to adapt to generate a new rdb by feeding it all the dats
My theory is that feeding it multiple dats should yield the desirable result, since any duplicate would just overwrite the previous one, that has to be tested though.

If you need help with that, let me know

markwkidd:
@larrykoubiak I just pushed two more files - `MAME - Consolidated ROM Sets.dat` and `MAME - Consolidated BIOS and Devices.dat`

Entries with matching zip filenames and matching SHA1s are not duplicated, but entries with the same filename but different SHA1s were both kept. I hope this is what we're looking for.

After a little while trying to recover some fluency in bash scripting, I wound up writing a simple AutoHotKey script to handle the consolidation instead, at least for now.

markwkidd:
@larrykoubiak I modified my script to check for hash collisions between the consolidated MAME dat and the FB Alpha (Split) DAT. There are a lot of identical hashes. :(

Here is a DAT with the list of entries that have identical CRC32 hashes: 
[crc32_mame_fbalpha(split)_hash_collisions.zip](https://github.com/libretro/libretro-database/files/783802/crc32_mame_fbalpha.split._hash_collisions.zip)

larrykoubiak:
@markwkidd i don't think that'll be an issue, since an rdb can't contain duplicates based on keys, it'll just overwrite the same data during generation, so you don't have to worry about that i think

markwkidd:
@larrykoubiak  Wouldn't that mean that it overwrites those entries in the FB Alpha database?

larrykoubiak:
@markwkidd it all depends on the dat files you pass to c_converter, if you only pass MAME dats when generating the rdb, it will only affect those. Now what i don't know is how the scanner determines which rdb to use in case of a duplicate entry across rdbs, but it'll probably pick the first match, would have to analyze [task_database](https://github.com/libretro/RetroArch/blob/master/tasks/task_database.c) to be sure

markwkidd:
@larrykoubiak OK, that makes sense. As you say the duplicate entries I've filtered could only be an issue across RDBs, not within c_converter.

Is there anything else I can do to help with this investigation at this point?

larrykoubiak:
well you can just generate the rdb with c_converter or better, adapt the
libretro-build-database.sh,
then make a test scan in RA, if it works, pull request and all done

markwkidd:
@larrykoubiak I generated an RDB for all of the MAME CRCs (with some help from @retro-wertz)

Here is the location of the new file in my fork if you want to try it out: https://github.com/markwkidd/libretro-database/blob/master/rdb/MAME.rdb

metadat/mame/MAME 0.139 Split TorrentZipped (No BIOS).dat
```diff
@@ -0,0 +1 @@
+
```
RobLoach: MAME 0.139 Split TorrentZipped (No Bios) empty?

RobLoach:
I like the idea of splitting the MAME into different versions for each of the different cores.

1. Could also update the .info files for those cores to target the correct database in libretro-super. Not sure that makes sense, since then you may find duplicates in the Menu though....
2. Don't think we need to add the XML files in this... They could live elsewhere.
3. Maybe we try targetting just the latest MAME? I really like the idea of supporting TorrentZip here. That seems to be the standard. Would we use both Split and Merged? Or just split?

retro-wertz:
split and merged mame rdb should be great and should start building for
0.183 sets(assuming @markwkidd has complete sets), but does that mean some
CRC will clash with FBAlpha?(cps1,neogeo and other older stable romsets)

markwkidd - have you already tested what retroarch will prioritize which
rdb to use when there is crc clash?

markwkidd:
@RobLoach I'd really like to see playlist support for one of the MAME cores available in Lakka (MAME 2003 and MAME 2014). MAME 2003 has the biggest current user base and runs on the most hardware if there can only be one Lakka MAME version targeted. 

I'd like to strongly advocate for supporting Non-Merged TorrentZipped sets, at least for the frozen MAME cores. That is the most common format for those versions. Current versions of MAME are most widely available as Split sets.

Merged sets are around but from what I can tell they are in third place in terms of availability and frontend support.

markwkidd:
There is no way to avoid hundreds of identical CRCs between MAME and FB Alpha whatever concept we com up with for a MAME database, even if we don't try to scan for every supported version of MAME and FB Alpha ROMs.

I'm going to try to keep helping move this current effort along because I'm eager to see what other folks thing, but I have started to believe that it would be better just to have one combined "Arcade" database that combines FB Alpha and MAME CRCs. Then let the user select which arcade emulator they want to use for games in the Arcade.lpl playlist via the RetroArch playlist settings menu. This would be more like the way console ROMs are scanned -- RA doesn't try to assign each SNES ROM to the best emulator for that game -- it just creates a SNES playlist and lets the user customize it.

At any rate, I'd be happy with anything that helps us Lakka users be able to generate some playlists that include our MAME games. :)

markwkidd:
@larrykoubiak have you had a chance to look at this work yet?

twinaphex:
Can we come to a consensus here on whether to merge this or not? @larrykoubiak @RobLoach etc.

markwkidd:
@twinaphex here's my take: it's not ready, and it's not going to work. I would earnestly welcome anyone who is willing to take the time to evaluate my work, even, or especially, if they disagree with this assessment.

I have provided the complete, and time-consuming (15-20 hours of work) raw materials for an investigation into whether the current scanning system can properly accommodate arcade ROMs. While I am qualified to provide these materials, I'm somewhat less qualified to make a definitive decision as to "nay". That would require an investment of time and energy from someone who knows the RetroArch DB system better than I'l be ever.

That said, I believe that if someone were to sink the 3-10 hours needed to evaluate my work, they would determine that a completely different mode of scanning is necessary for the RetroArch playlist scanner to support arcade ROMs.

retro-wertz:
an RDB can have different crc for a game title, but a title having the same crc or different rdb, just as the case for FBA and different versions of mame, not sure which rdb would be used.

markwkidd:
Everything that I submitted is TorrentZipped.

RobLoach:
Built the database after applying the attached patch to libretro-super:
[mame.patch.txt](https://github.com/libretro/libretro-database/files/1079874/mame.patch.txt)

The built database is at:
http://files.robloach.net/MAME.rdb

I'm conflicted on supporting older versions. Some of the entries match across multiple versions, but that may be a good thing for MAME. We opted to not support old versions of No-Intro, and other DATs, but MAME may be an exception, at least for the existing cores we have for it.

Issue #243, opened 2016-08-15
GoodTools


RobLoach:
Add the GoodTools DATs. Need to convert from the Tools, to a DAT. Anyone willing to do that?

zandrorev:
From the license.txt:

>   You are NOT authorized to reverse engineer or otherwise extract the
>   database of names included inside the GoodTools. This includes 
>   generating databases for other renaming programs.

I suggest you look into the MAME Software Lists (formerly managed under the MESS banner), as there is significant crossover in provided game information, it's more fleshed out meta-wise, and they had the GoodTools author's support in developing some of their dats.  But do check their license for potential usage restrictions.

RobLoach:
Agreed. GoodTools is quite old, and we don't want to go against the license.

PR #431, opened 2017-03-06
Added some goodtools metadats and updated rdb

JuanVCS:
Should help people not finding their roms with the scanner and they also have some translations/romhacks.


hizzlekizzle:
I'm not 100% but I thought we couldn't include GoodTools sets due to licensing...?

JuanVCS:
I'm adding a few more sets to the already existing https://github.com/libretro/libretro-database/tree/master/metadat/goodtools

RobLoach:
I'm not a license lawyer, but to my knowledge, it's the GoodTools binaries and source files that we cannot host. Looking at the [Hosting License](https://en.wikipedia.org/wiki/GoodTools#Hosting_license), that's what I come to understand. GoodTools is a set of tools, these DAT files are files that JuanVCS built, so they should be good to go.

Is anything missing before we merge this, @JuanVCS?

JuanVCS:
This is all I have at the moment so I guess they're good to go. BTW they were built with clrmamepro's dir2dat utility, no goodtools involved.

hizzlekizzle:
Quoted from #243 :

> You are NOT authorized to reverse engineer or otherwise extract the
> database of names included inside the GoodTools. This includes
> generating databases for other renaming programs.

This sounds like we're not allowed to do much of anything with them. However, this seems like a pretty broad claim and I don't know how enforceable and/or legally legitimate it is.

RobLoach:
Hmmmm, we should likely remove them then. Thoughts?

twinaphex:
Do other programs disregard this stipulation inside the license?

zandrorev:
The GoodTools use various methods to determine whether roms of varying structure match an entry. This includes looking past the sporadic inclusion of arbitrary headers in many sets, accounting for byte order or interleaving of N64, MSG/MGX format for NEC PCE, and so on.  Despite these checks, the GoodTools will not modify the roms to complement its internal DB, for in an emulator there could be dependence on instructions in a rom's header.

A dir2dat would capture a collector's (or his source's) unique form of a set, and wouldn't always represent somebody else's accurately.  We have been battling GoodSet dir2dats for years, not just because of licenses, but also because they create conflict and misunderstanding from those who are not sufficiently aware of the dynamics of the GoodSets.

ClrMamePro was equipped with header checker/skipper functionality which can be used to mimic some of the the GoodTools methods, but not all, and only leads to the same results of giving leniency to differing sets rather than assuring strict conformity (which would break compatibility).

I have to recommend against incorporating dir2dat forms of the GoodTools sets, especially since they are not fully reliable portrayals of them, unlike No-Intro dats, which provides hashes of headerless roms and otherwise aims to adhere to predetermined parameters.

PR #435, opened 2017-04-18
Update No-Intro DATs

RobLoach:
No-Intro has deprecated the use of the legacy clrmamepro format in favour of XML. [libretro-dats](https://github.com/RobLoach/libretro-dats) can port the XML over to our old DAT format.

This PR is the result of running through https://github.com/RobLoach/libretro-dats/pull/11 .


twinaphex:
Glad we can avoid the XML files.

PR #438, opened 2017-05-18
Remove Pokemon Yellow (GB overlap) and rdbs rebuilt

JuanVCS:
Else games would show up twice on both GB and GBC playlists.


RobLoach:
Looks like it is in the meta data:
https://github.com/libretro/libretro-database/search?utf8=%E2%9C%93&q=964b7a10&type=

JuanVCS:
Problem was it being listed as a Gameboy Color game as well, which I edited out of that specific dat.

PR #460, opened 2017-06-17
Add MAME DATs (Non-Merged + Split)

RobLoach:
@markwkidd This is looking great! In order to allow building the different databases so cores can load different RDBs correctly, I took on a few things...

1. Removed `TorrentZipped` in the names
2. Removed `and Devices`
3. Removed `No BIOS` note
4. Moved them into `mame-split` and `mame-nonmerged` folders
5. Updated libretro-super to allow building the databases at https://github.com/libretro/libretro-super/pull/471

@markwkidd Have a couple questions for you...

1. We could have a `mame-nonmerged/MAME.dat` and `mame-split/MAME.dat` for the latest and greatest and not have to consolidate them manually. Thoughts? Maybe next update we do that.
2. The libretro-super update is over at https://github.com/libretro/libretro-super/pull/471 , mind having a look?
3. Should we add a `mame-merged` directory? Does RA allow launching games directory from a merged file?

Looking nice! Thanks a lot for going through this, must've taken a while to build the DATs.


markwkidd:
This is looking great @RobLoach 

> We could have a mame-nonmerged/MAME.dat and mame-split/MAME.dat for the latest and greatest and not have to consolidate them manually. Thoughts? Maybe next update we do that.

Yes I think that makes the most sense. I don't have MAME sets built for MAME 0.174 and the latest version, but if this overall approach is working I'd try to contribute those in a few weeks.

> The libretro-super update is over at libretro/libretro-super#471 , mind having a look?

Looks great!

> Should we add a mame-merged directory? Does RA allow launching games directory from a merged file?

My understanding, although untested, is that you can launch MAME clones from merged sets in RA by renaming the file to be the filename of the clone you want to load. Maybe something could be done with symbolic links, but honestly Merged sets (and split sets too IMO) are a gimmick unique to the MAME universe. Is it worth it? :)

markwkidd:
Hi @RobLoach I noticed that no .rdb is being generated for MAME 0.78

RobLoach:
Thanks, fixed at https://github.com/libretro/libretro-super/pull/471/commits/fcf7a16e0019f51334afe130094377e6c53d7dc0 .

markwkidd:
@RobLoach I'm looking at this URL and still not seeing a MAME 0.78 RDB DAT: https://github.com/libretro/libretro-database/tree/master/rdb

I'm also having trouble scanning for a ROM that should be in that database. Am I looking in the right place, or misunderstanding something?

RobLoach:
@markwkidd The core .info files are not merged yet. If you bring https://github.com/RobLoach/libretro-super/tree/mame-versions/dist/info into your local core info directory, it should register the new RDBs.

Could do with another update of these RDB files too.

PR #481, opened 2017-07-13
Update FB Alpha .dat and .rdb to 0.29.97.40. Fixes #329.

### LINKED ISSUE #329: [FB Alpha] Update to 0.2.97.39 and TorrentZip
this database(and probably others) really needs to be updated. its still based on 0.2.97.38 (torrentzipped) and the core for windows has already been updated to 0.2.97.39 meaning my updated cps2 roms cannot be imported to playlist which the reason why i like to use retroarch. 

also just for fun i made mame.78 neogeo romsets verified with clrmame-torrentzipped and crc never matches the ones in mame2003.rdb list aside from one rom, shocktra.zip and placed it FB Alpha - Arcade Games. i was in the impression that one purpose of rdb is for playlist names.

either the rdb should be updated to reflect core versions used in repository, or probably new a different method of scrapping zipped romsets in RA since it takes a long time scanning each file in the zip anyways when its rdb is only checking for the zips crc.

SpiralBrad:


RobLoach:
Thanks a lot! :+1:

PR #482, opened 2017-07-14
Add FBA_dat_gen.py script

SpiralBrad:


RobLoach:
Thanks a lot, this is handy. Did you use this on the last dat you made?

SpiralBrad:
Yep! Works a charm. Uses the FB Alpha generated dat for the textual metadata and a provided romset for zip sizes and hashes.

PR #485, opened 2017-07-16
Final Fantasy V Cheats

brianblakely:
I noticed that this was the only FF for SNES that didn't have any cheats in RetroArch.


RobLoach:
Nice work!

PR #488, opened 2017-07-18
BIOS: Add mising MD5 hashes and update MSX

Bracket-:
First commit adds missing MD5 hashes

Second commit updates MSX's MSX.ROM & MSX2P.ROM hashes to the latest provided by fMSX.

From what I understand, the bios files are for fMSX since blueMSX uses C-BIOS so it makes sense to use the hashes from the bios files that fMSX provides. The other MSX bios hashes also match the bios files included by fMSX.


RobLoach:
Thanks!

PR #497, opened 2017-08-01
Update Final Burn Alpha DAT/RDB to 0.2.97.42

SpiralBrad:
- Fixes some issues that remained in the last update for dmnfrntpcb.zip and svcpcb.zip; the FBA-generated ClrMAMEpro dat reports these as not needing the PGM BIOS roms, but said zips with the missing files will not run. Generated with these corrected this time, so this dat/rdb works for those now. (Tested and working.)
- DAT/RDB created with a set from a certain *dome.


RobLoach:
Nice work on the metadata, too! :+1:

SpiralBrad:
That's all the FBA_dat_gen.py script. ;) - FBA team updates entries in their database, script grabs it from the FBA-generated ClrMAMEpro dat. Quick & easy.

PR #516, opened 2017-09-17
Update databases to use primary data tracks instead of cue/gdi files

bkoropoff:
Companion to https://github.com/libretro/RetroArch/pull/5447


RobLoach:
Nice work! :+1:

PR #524, opened 2017-09-24
GoodSets created using dir2dat

xTMODx:
i am not sure about the dat name if they not correct please correct them


RobLoach:
Nice work, and yes, the filenames should match those of the Database names.

xTMODx:
Ok so GoodNES is ok then and not Nintendo - Nintendo Entertainment System like for No-Intro?

RobLoach:
Should match those of the No-Intro filenames:
https://github.com/libretro/libretro-database/tree/master/metadat/no-intro

xTMODx:
Do they have to match for a reason? 
Should i split GoodGBx to 

"Nintendo - Game Boy" & "Nintendo - Game Boy Color"
or can i rename to
"Nintendo - Game Boy - Game Boy Color"

same would be for GoodWSx...

retro-wertz:
name, description, rom labels? with name and description the same entry? never realised our database is like this now. this used to be name, publisher, rom.

Anyways is these romsets, especially for snes, nes famicom disk system, verified that the headers are correct? im asking coz since the addition of the goodnes set there has always been missing entries during the import due to wrong headers, and i have verified most of the goodnes sets that i have.  even the no-intro based are missing something about 200-300 stuff.

RobLoach:
> Do they have to match for a reason?

When building the RDB files, the database builder merges all .dat files of the same name. GoodGBx should likely just be `Nintendo - Game Boy Color.dat`

RobLoach:
Going to merge to `goodtools`, and then fix it there.

PR #538, opened 2017-10-13
Remove GoodGBx

RobLoach:


Kivutar:
Yeah good move, thanks. This was causing sorting bugs with GBC.

PR #564, opened 2018-01-06
Create Mattel - Intellivision.dat from TOSEC

markwkidd:
Source: Mattel Intellivision - Games (TOSEC-v2014-01-18_CM).dat

Source DAT download: https://www.tosecdev.org/downloads/category/41-2017-11-01


RobLoach:
Here ya go! https://github.com/libretro/libretro-database/pull/565/files

markwkidd:
Thanks! It's been a while since my MAME DAT work and I guess I misremembered the process.

RobLoach:
MAME is different than the TOSEC set. May be almost time to update those MAME ones too!!!

markwkidd:
There is definitely some action in MAME 2003 DAT land. There are a few of us tracking those changes, although I don't think the dust has settled on exactly which ROMs are going to work with the enhancements. I'll totally guess that the MAME 2003 core's DAT will be stable again in another month.

Are there other MAME DATs that need an update?

RobLoach:
Unsure, haven't used them.

PR #567, opened 2018-01-07
Fix database regression

bkoropoff:
Use primary data track CRC from redump dats rather than the first track.  This was lost somehow during an update.

cc @RobLoach @twinaphex


metadat/redump/NEC - PC Engine CD - TurboGrafx-CD.dat
```diff
@@ -5,40 +5,94 @@ clrmamepro (
...
+	name "1552 Tenka Tairan (Japan)"
+	description "1552 Tenka Tairan (Japan)"
+	rom ( name "1552 Tenka Tairan (Japan) (Track 02).bin" size 33694752 crc 160d1887 md5 36fbc21bbbfda636ba57790081f6a459 sha1 7873f67517748b1a71a18f12db2e6f204e7f5c61 )
+)
+
 game (
 	name "3 Games in 1 - Gate of Thunder + Bonk's Adventure + Bonk's Revenge (USA)"
 	description "3 Games in 1 - Gate of Thunder + Bonk's Adventure + Bonk's Revenge (USA)"
-	rom ( name "3 Games in 1 - Gate of Thunder + Bonk's Adventure + Bonk's Revenge (USA) (Track 01).bin" size 7916832 crc ecf8b16b md5 e25cf5e45442703396a239611ea901be sha1 9a8ba8d225d9575ab09b8a612cb86164c4902cac )
+	rom ( name "3 Games in 1 - Gate of Thunder + Bonk's Adventure + Bonk's Revenge (USA) (Track 22).bin" size 10527552 crc b81377a7 md5 2554cd8a8fdbfd258ec36250a7cf93c0 sha1 6289ffce85eea8add87481bf774c9c1f2ec78939 )
```
RobLoach: Looks like it's changing from the first track. How do we know which track is the primary track?

bkoropoff: The primary track is the largest data track.  This change was made back here:

https://github.com/RobLoach/libretro-dats/pull/20

RetroArch uses the same heuristic to decide which CRC to query.

RobLoach:
These are built through https://github.com/robloach/libretro-dats . Would be cool to figure out which track is primary.

metadat/redump/NEC - PC Engine CD - TurboGrafx-CD.dat
bkoropoff: This comment mentions using the largest data track rather than the first to avoid collisions: https://github.com/RobLoach/libretro-dats/pull/20#issuecomment-330095231

bkoropoff:
libretro-dats should be using the largest-data-track heuristic to determine which is primary.  I just reran it in order to create this commit.  Not sure how the regression snuck in.

RobLoach:
Great! Thanks.... So, it'll go through all tracks, and pick the LARGEST track on there. The logic should likely be updated around here:
https://github.com/RobLoach/libretro-dats/blob/master/index.js#L208-L232

Thanks for finding that! If we merge this and thne update libretro-dats, we should make sure the regression doesn't happen. Could you double check the logic?

RobLoach:
Rebuilding this, I get Track 1's. Do you have any changes that aren't pushed up in libretro-dats?

bkoropoff:
@RobLoach 

I'm on d39d27cfbd72b7, which is the tip of master

PR #570, opened 2018-01-25
docs: Add some build docs

RobLoach:
@markwkidd Is this helpful?
https://github.com/libretro/libretro-database/blob/docs-build/README.md#building


markwkidd:
@RobLoach looks good, thank you! Now that I see your instructions I'm pretty sure I've done this before myself in the past but it was definitely not going to come back to me!

PR #580, opened 2018-02-21
Remove GoodSNES

RobLoach:
https://github.com/libretro/RetroArch/issues/6321


RobLoach:
Which are considered super-sets?

i30817:
Trurip takes from TOSEC if iirc (i suspect they don't want the divided tracks approach redump takes). 

No-intro inherited a lot of roms from goodsnes and goodnes.

It's not a question of 'who takes from whom' but 'which set is saner' or 'which set does the user want'.

PR #591, opened 2018-03-19
Rename SNES Hacks to SNES

### LINKED ISSUE #590: Hacks
This file https://github.com/libretro/libretro-database/blob/db65621603847df1dc0e27b0f5d98c61a6be49cf/metadat/hacks/Nintendo%20-%20Super%20Nintendo%20Entertainment%20System%20Hacks.dat

should be renamed without the `Hacks` in the end.

If not, a separate playlist will be created

RobLoach:
Fixes #590


twinaphex:
No, why are you doing this again? I do NOT want this to be part of the main Super Nintendo database.

At least explain the logic behind this, all of these hacks clutter up the entire list and make it impossible to find any legitimate entries.

RobLoach:
There are only 6 entries here, and it likely won't grow much from there. We could delete it entirely, but it's just 6 at this point.

Merging, and then we can reconsider the direction here if it grows more.

PR #603, opened 2018-04-08
SNES: Restore GoodSNES

### LINKED ISSUE #598: Virtual Console entries no longer found
I noticed with the latest version of the database that a bunch of Virtual Console entries weren't getting recognized by RetroArch's scanner - looking at the current revision I'm not entirely sure where these would be located; I thought that it might be this issue I found when building the Wii database...

```
/home/steam/libretro-super/retroarch/media/libretrodb/dat/Nintendo - Wii.dat:66349:6: fatal error: Unexpected '(' instead of key
```

I'm assuming this is a red herring though, as I'm not using the full WADs, just the extracted ROM with sound restored (so previously the files were recognized as "(VC) Sound Restored"). Files that are the same as the original SNES data are fine (Star Fox 1/2, Yoshi's Island). The last time this worked was with the SteamOS-Tools version, v1.6.1 with a couple patches for SDL_GameController support:

https://github.com/mdeguzis/SteamOS-Tools

Is there a separate database I should be looking at, or is there a bug I should track down?

RobLoach:
Fixes #598 cc @flibitijibibo


i30817:
Won't this cause scans of SNES 'no-intro' roms to start getting goodsnes names? Because as was said, plenty of dumping projects don't do anything to the rom except rename it.

This could be solved by mandating the users to rename their roms and checking for both checksum and name on the RA scanner...

RobLoach:
If there are duplicates, it'll get NoIntro names. Check ordering in libretro-build-database.

i30817:
People start complaining when a few roms don't have duplicates and match the second dump and ruins their 'aesthetics'. 

I guess it's not very relevant to the database but RA seems awfully blase and strict at the same time about what ROMs it will scan or pair up with features like savestate netplay. Wii games get recognized by serial, and can get a wrong checksum beacause the entry picked up never actually checked for the crc32/md5sum and then (probably) get a scrubbed and unscrubbed iso paired up in netplay.

When CRCs match, a goodsnes style game can get 'renamed' to a no-intro style, or a goodsnes style game can be alone on a bunch of 'no-intro' named games that were renamed by a ordering accident.

PR #614, opened 2018-04-11
Add Saturn hacks file

i30817:
RFC about this i guess. Before trying something heavier like the psx1 hacks i have i decided to limit myself to the few Saturn hacks.

Things i don't understand:

The 'rom name', how does it behave when the rom is multifile and the file with the right CRC is not the 'index' file (that honor would go to cues).

Because Mednafen for one will not load games that don't have a cue (even bare isos in MODE1/2048 mode) but will load them fine if passed the right cue.

So my hope is that the Scanner finds cues, finds the first track, tries to check for serial, fails, tries to check for crc, finds it and then adds the cue to the playlist.

Not  Scanner finds cues, finds the first track, tries to check for serial, fails, tries to check for crc, finds it and then adds the *track* to the playlist.

Even if it works like i expect, if a serial method is created for saturn suddenly these hacks wouldn't get scanned because the Scanner would think it had 'found' the game, in spite of it being a hack. So how is this supposed to work,?

Also, can i use comments like this?

### DIFF
--- a/metadat/hacks/Sega - Saturn.dat
+++ b/metadat/hacks/Sega - Saturn.dat
@@ -1,0 +1,47 @@
+clrmamepro (
+	name "Saturn hacks"
+	description "Romhacks of Saturn games"
+)
+
+game (
+	name "Dragon Force II: The Godforsaken Land"
+	description "Translation by Verve Fanworks version 1.04"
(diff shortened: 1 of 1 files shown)


RobLoach:
Looks good! Thanks. Next DB update, these should likely appear.

PR #616, opened 2018-04-12
add many N64 hacks and translations

i30817:
Many of these are because N64 emulators unfortunately accept byteswapped, little endian and big endian byte orders, and no-intro and goodn64 (and gamecube iirc) use different byte orders (or worse, many).

Even with this this is severely incomplete on some things. For instance, hack patches versions of the 'GameCube to N64' series of which the last 3 games added here are part have patches for

3 zelda games (OoT, OoTMQ, MJ) * 3 regions * 2 (with aliasing or not) * 3 (bytes orders) = 54

All things considered i decided to only add the main 3 games on the USA with the anti-aliasing disable patch (my preference) and in the byte order the patch expects (big endian). Which is not btw, the order of no-intro, so to make a small soft-patch to no-intro, you'd have to patch the big-endian rom, convert it back to no-intro byte-swapped and make the bps patch finally.

Of special notice is `Sin and Punishment: Successor of the Earth (USA) (Wii U Virtual Console)` which is a rom that needs to be extracted from a Wii U wad to work on n64 emulators but i still provided all the byte orders because i created a softpatch for the original N64 Japan release using that method. If RA ever gets VFS softpatch support those different byte orders might allow me to scan softpatches and get the hack definition, not the 'original game' like currently.

In fact this is kind of pointless to me, because i created softpatches for everything but cds so everything doesn't get picked up until VFS soft-patch happens, but ....

### DIFF
--- a/metadat/hacks/Nintendo - Nintendo 64.dat
+++ b/metadat/hacks/Nintendo - Nintendo 64.dat
@@ -9,3 +9,127 @@ game (
 	rom ( name "SM64 Last Impact V1.1.z64" size 67108864 crc 4ce7fa2a md5 836667a37dce62414064b62f5dafe9f6 sha1 78e3e86f679330a40a0480372f65c362179042ff )
 )
 
+game (
+	name "Custom Robo (Japan) [T-En by Star Trinket]"
+	description "'Custom Robo' translation by Star Trinket version 1.0"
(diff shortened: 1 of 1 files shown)


RobLoach:
Also note that soft-patching works for some of these too :+1:  ...

https://docs.libretro.com/guides/softpatching/

Looks like it doesn't work for N64? Hmmmm.

Anyway, this looks great! I'll review soon.

i30817:
Aroenal just put in a update in the romhacking queue so maybe i'll be able to get a saner patch link and version for the zelda roms.

PR #617, opened 2018-04-15
add many nes hacks and translations

i30817:
This was done by taking advantage of my OCD and with a script i made and with a compilation of flips to run on linux (script may or may not run on windows with flips.exe on the same directory, haven't tried).

Since the very beginning i started using hacks i've softpatched, and lately i've been making all my hacks readme have a version in the filename of the readme, and a link to the hack on romhacking at the top of the readme.

So i made a tool that extracts local version and url, goes to the hack page, scrapes the hack name, author, version and some odds and ends, patches locally the game with the current ips/bps on the dir to a tmp file, calculates size, crc, md5, sha1 of the resulting hack and creates the entries on the dat file and warns if the hack needs updating (because of different versions).

Might be helpful to make listings from stuff that's on romhacking and applies cleanly (there is some stuff that doesn't).

A warning: although this calculates the right checksum values on the where you combined two hacks into a single softpatch file, you'll end up with 'only' the first hack link on the readme and the name of that hack. On these files, the only one that had that problem was `Castlevania 3: Improved Controls + Localization / Title Screen Redone` which i edited manually (i added the '+ Localization / Title Screen Redone' which is the second hack) for example.

I also didn't allow to add romhacks that flips said failed the checksum so some hacks won't be here (besides the ones that i never downloaded anyway). One example of such is `Castlevania: Chorus of Mysteries`, which is supposed to apply to some janky rom that doesn't seem to be on the net anymore, since every rom i tried flips complained (and it's a ips patch to boot).

It should be possible to apply this script to the SNES and other collections, I'll open a PR for snes in a few days if no requests to the script are made here.

flips is here: https://github.com/Alcaro/Flips and can be built in cmd line mode with a makefile switch `make CLI=1`

script is here (python 3).
[makedat.py.txt](https://github.com/libretro/libretro-database/files/1911638/makedat.py.txt)

### DIFF
--- a/metadat/hacks/Nintendo - Nintendo Entertainment System.dat
+++ b/metadat/hacks/Nintendo - Nintendo Entertainment System.dat
@@ -33,3 +33,2158 @@ game (
 	rom ( name "Sonic_NES_Improvement_plus_Music_plus_Graphics.nes" size 786448 crc a38eceeb md5 0e2214993648014f95544b310ee6a11d sha1 b8a13b58d52d13fe1e616bb1c4fff6f827dc5890 )
 )
 
+game(
+    name "Cosmo Police Galivan (Japan) [T-En by Jair]"
+    description "Translation of 'Cosmo Police Galivan (Japan)' by Jair version 1.00"
+    rom ( name "Cosmo Police Galivan (Japan).nes" size 262160 crc 4f031cb0 md5 08ecfd17a791acde0bf63543c98115a6 sha1 fd3d9f8a6c862773194c3e23b420aabe9453d2d6 )
+    comment "https://www.romhacking.net/translations/109/"
+)
+
+game(
+    name "Kamen no Ninja - Hanamaru (Japan) [T-En by pacnsacdave]"
+    description "Translation of 'Kamen no Ninja - Hanamaru (Japan)' by pacnsacdave version 1.0"
+    rom ( name "Kamen no Ninja - Hanamaru (Japan).nes" size 262160 crc c4924b42 md5 3858f554bb79f8522c45b57b997f80a5 sha1 93e5757e77bf6588a8723360e9396cd6292df412 )
+    comment "http://www.romhacking.net/translations/2331/"
+)
+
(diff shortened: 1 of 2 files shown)


retro-wertz:
great... more hacks...

i30817:
BTW, i can change the titles somewhat but romhacking is sort of inconsistent about them.

For translations i'm taking [the first title on the side column 'translation information'](http://www.romhacking.net/translations/3544/) for the 'main name and add the [T-Language by author]

These first titles are what the game or hack is commonly known as (at the time of creation of the page so it might be very different from Nointro names for example).

For translations the main name could be instead:

1. Me using clrmamepro to rename the names that don't match (hacks because i renamed the roms) to no-intro format and use the filename always. That would be consistent for translations but bad for hacks (that have their own names). I can distinguish between them in code so no big problem about that. 
2. Another way would be to use the first of the alternative titles (below the title screen on romhacking. Sometimes these have [nice translations of what the game is really about](http://www.romhacking.net/translations/3544/), sometimes [they have a place holder](http://www.romhacking.net/translations/3545/) that could be recognized and changed back to the main title.

I'm currently using the main title for hacks, because the other strategies wouldn't capture the name the hack author decided to give it.
For hacks the 'secondary' title part of the [Hack information column](http://www.romhacking.net/hacks/3964/) shows what the author decided to name the original rom, which might not be consistent either.
There is the possibility of (again) using no-intro names but for some roms i changed the soft-patches so they would work on no-intro (by patching the rom the author wanted and then creating a patch that goes from the no-intro rom to the patched rom) so the names might actually be even more misleading (if pretty).

What's your opinion about these two choices?

i30817:
Uhh, btw for the NES do i have to skip the iNES header during the calculation of crc, md5, sha1 in order for RA to recognize these checksums?

I jut remembered that that was a thing with redump and hacks wanting roms from with the headers.

i30817:
Opps, i screwed up editing the files. I'll close this one and open another.

PR #619, opened 2018-04-15
add many nes hacks and translations

i30817:
Same as the closed PR.

I verified that RA checksums don't skip the header like redump dats do so the output rom() entry checksums are of the whole patched file.

This was made by 
having a cli version of flips on the same dir as the script (built with `make CLI=1`)
having a no-intro (xml format) dat for nes to pass (to use the no-intro names when possible instead of the names on the romhacking page)
and using this script:
edit, script updated. edit: twice:
https://gist.github.com/i30817/c5332248f46113fcb4ca03081f7673f2

with the following arguments:
`./makedat.py NES/ nes -d no-intro-nes.dat`

This verifies the games against the no-intro dat and if it finds it, use that name, but if it doesn't, warn use the romhacking.net page name. In effect this only 'warned' on unlicensed games translations (since i had no hacks for unlicensed games).

Hacks that aren't translations use the name on the redump page always but its extended description might use the dat name if it's found because those are more likely to match the local rom name than the romhacking entry. But if not found still use romhacking 'secondary' title.  I haven't a example of this because i've not seen a hack that isn't a translation of unlicensed games on my collection.

A warning: although this calculates the right checksum values on the where you combined two hacks into a single softpatch file, you'll end up with 'only' the first hack link on the readme and the name of that hack. 
On these files, the only one that had that problem was `Castlevania 3: Improved Controls + Localization / Title Screen Redone [Hack by NaOH and ShadowOne333]` which i edited manually.

I also didn't allow to add romhacks that flips said failed the checksum when applying the rom so some hacks won't be here (besides the ones that i never downloaded anyway). One example of such is Castlevania: Chorus of Mysteries, which is supposed to apply to some janky rom that doesn't seem to be on the net anymore.

### DIFF
--- a/metadat/hacks/Nintendo - Nintendo Entertainment System.dat
+++ b/metadat/hacks/Nintendo - Nintendo Entertainment System.dat
@@ -33,3 +33,2214 @@ game (
 	rom ( name "Sonic_NES_Improvement_plus_Music_plus_Graphics.nes" size 786448 crc a38eceeb md5 0e2214993648014f95544b310ee6a11d sha1 b8a13b58d52d13fe1e616bb1c4fff6f827dc5890 )
 )
 
+game (
+    name "Cosmo Police Galivan (Japan) [T-En by Jair]"
+    description "English translation by Jair version (1.00)"
+    rom ( name "Cosmo Police Galivan (Japan).nes" size 262160 crc 4f031cb0 md5 08ecfd17a791acde0bf63543c98115a6 sha1 fd3d9f8a6c862773194c3e23b420aabe9453d2d6 )
+    comment "https://www.romhacking.net/translations/109/"
+)
+
+game (
+    name "Kamen no Ninja - Hanamaru (Japan) [T-En by pacnsacdave]"
+    description "English translation by pacnsacdave version (1.0)"
+    rom ( name "Kamen no Ninja - Hanamaru (Japan).nes" size 262160 crc c4924b42 md5 3858f554bb79f8522c45b57b997f80a5 sha1 93e5757e77bf6588a8723360e9396cd6292df412 )
+    comment "http://www.romhacking.net/translations/2331/"
+)
+
+game (
+    name "J.League Winning Goal (Japan) [T-En by Stardust Crusaders]"
+    description "English translation by Stardust Crusaders version (1.0)"
+    rom ( name "J.League Winning Goal (Japan).nes" size 262160 crc fbfa06f6 md5 c357a2367d3f13c448479731df3c54fe sha1 22aaa0c031b9db96684d5a37a81192986abd7b5d )
+    comment "http://www.romhacking.net/translations/1363/"
+)
+
+game (
+    name "StarTropics - No Movement Delay [Hack by Trisma + 1 hacks]"
+    description "StarTropics - No Movement Delay hack by Trisma version (1.0) + StarTropics (Music Fix) hack by rainwarrior version (1.0)"
+    rom ( name "StarTropics (USA).nes" size 524304 crc 64b8860f md5 0d13248466eea1dedb6131378780e51d sha1 447d8196162dd873b67c4cbe06db690078440a02 )
+    comment "http://www.romhacking.net/hacks/3307/"
+    comment "http://www.romhacking.net/hacks/2229/"
+)
+
+game (
+    name "Daisenryaku (Japan) [T-En by Aeon Genesis]"
+    description "English translation by Aeon Genesis version (1.0)"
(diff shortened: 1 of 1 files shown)


i30817:
please don't merge this yet, i have a better version of the script about to work that handles multiple patches on the descriptions. The script is also handy for finding updates to patches you have...

i30817:
Ok, can be review i guess. The rest is a question of quote this, separate that. Right now the script can support multiple links in a 'version' file and adds info about all of the patches. There is a complicated algorithm on the script but it basically has this idea:

        #main name priority:
        # first hack with zero translations > rom_title > last translation > first hack
        # secondary name be:
        # [(last)T-lang by author] for 1+ T and 0 hacks (translations are only multiple if addendums)
        # [(last)T-lang by author + # Hacks] for 1+ T and 1+ hacks + 
        # [(first) Hack by author] for 0 T and 1 hack
        # [(first) Hack by author + # hacks] for 0 T and 1+ hack
        # (# is always equal to (number of translations + number of hacks -1) when it appears

And the typical output is:
```
game(
    name "Lost Vikings II - always stereo [Hack by sluffy + 1 hacks]"
    description "Lost Vikings II - always stereo hack by sluffy version (1.0) + Lost Vikings II - password saving hack by sluffy version (1.0)"
    rom ( name "Lost Vikings 2 (USA).sfc" size 1048576 crc 683eeea8 md5 fd2b2810f31299f3ebf070e7062614b2 sha1 296a418b12958e5432ea6926376ce810450de69f )
    comment "http://www.romhacking.net/hacks/3782/"
    comment "http://www.romhacking.net/hacks/3783/"
)
```

I'll open a snes one since it's so easy to parse if you have the version files.

RobLoach:
There are a large amount of entries here... Could we just bring in the top 100 or so? The larger we make the database, the slower scanning will be.

i30817:
There is a large amount of translated roms over the decades. I'm against cutting it. And i wouldn't know how to prioritize even if i wanted.

Notice that these are only the roms i have (i didn't find a secret romhacking dat), so there are even more floating around. Conservatively, a lot more.

retro-wertz:
to be honest, my previous comment(the other PR) was sarcastic... there are too many hacks/translations and even adding them all wont make it a guarantee that everyone will be able to have the same crc since nes headers can have extra bits (even on the needed 1st 7 bytes) that affects crc. for example mirroring can be horizontal or vertical for mapper-controlled games.

retroarch should support skipping headers for nes/fds/snes(probably other consoles too) when scanning/calculating crc.

RobLoach:
I'd also REALLY like https://github.com/libretro/RetroArch/issues/2033 as an alternative solution around this. Just allow people to add the hacks without even needing them in the database.

i30817:
the whole point of having the database is having the metadata. Cutting off a large amount of data because 'no this is too much' is already what RA does bad. I don't have anything against skipping headers, but you should first do that work on the scanner on Ra if you want it to happen. It's per platform too.

RobLoach:
> Cutting off a large amount of data the users use because 'no this is too much' is already what RA does badly.

The issue is scanning performance... The larger we make the database, the slower scanning gets.

i30817:
Doesn't sound like it's a big problem to make it a different db in that case that you need to download from the downloader.

But, i'm still not convinced about the performance argument simply because what really hurts there is scanning discs and cds that are unzipped and thus need to calculate crc32. Do you pass on the size of the file when you scan as a early bail out?

i30817:
Let me make my point a little clearer: people will have these files on the roms folders already and when they request 'scan here' to RA, it will have no chance but to scan. If they aren't found it just makes the scan (infinitesimally) longer because they'd have to search the whole database just to fail.

In fact that's one of the ways you could help speed up the Scanner. Order the database by filesize when built and create a binary search to the 'Rom to be scanned'. Only calculate the CRC of the file if there is at least 1 match to the exact byte size and search upwards and downwards of the first match until there is a match.

RobLoach:
I appreciate these additions but I think there are two issues....

1. Adding things many entries will add bloat and slow down the scanner
2. Like @retro-wertz mentioned, when someone applies the patch to their NES rom, it is not guarenteed to result in the same CRC, since the headers are not skipped when scanning in RetroArch.

The other PRs are looking great though. I'll go through and do some reviews.

i30817:
2. i used the sane defaults of not using headers, which is completely possible because of ipsbehead.

1. i already explained why i think that the speed argument is bogus. Simply said, (again) the actual speed hit will happen *if the user has the roms on the directory they scan* not because they're on the database. Prior to matching they have to calculate the CRC and that happens *regardless* of if the roms are on the database or not, because to even find them they have to scan.

Which means any 'speed' lost would be better regained on other places. Namely making crc32 calculation faster would be better, or adopting a two stage strategy for cds of 'only calculate crc32 if you need it instead of serials' (eg: there is a hack with the exact size of the rom for a simple strategy).

PR #620, opened 2018-04-16
add some snes hacks and translations in romhacking.net

i30817:
More of the same. Again made with the help of flips, nointro dat an this script 
https://gist.github.com/i30817/c5332248f46113fcb4ca03081f7673f2

### DIFF
--- a/metadat/hacks/Nintendo - Super Nintendo Entertainment System.dat
+++ b/metadat/hacks/Nintendo - Super Nintendo Entertainment System.dat
@@ -16,6 +16,12 @@ game (
 )
 
 game (
+	name "Chrono Trigger - Crimson Echoes (USA)"
+	description "'Chrono Trigger' hack by Kajam Laboratories"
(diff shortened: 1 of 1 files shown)


i30817:
BTW, for the snes i should mention that these are all un-headered patches even if the original was for a header. There is a tool called ipsbehead that can remove the requirement of a header from a ips patch, or you can simply remove the header after translation is done.

Because of this, it wouldn't be surprising that in the future people complaining about not finding a hack that is there. But you shouldn't replace the entry but add a new one.

i30817:
speaking of that let me update so Crimson Echoes unheadered works... done

i30817:
That would basically make this sort of useless because for every 1 that got recognized, there would be 14 that weren't. No thanks, i prefer to retry the PR when you convince yourself that the database size is not the problem while scanning and that more efficient ways to calculate crc (if not on zip files) or query the location of a entry (if so) would be much better.

Of the people who complain about the scanning performance how many of them didn't have zips and complained before the crc method for large cds? Not that i don't want the crc method to be selectively applied again with a pre-filter for file size, but currently it's already much better than it was a few months ago in the degenerate case.

i30817:
Closing this until there is a Scanner rework that makes it faster.

PR #621, opened 2018-04-16
add new genesis hacks

i30817:
Hacks for megadrive, all uptodate.
Made with [the script](https://gist.github.com/i30817/c5332248f46113fcb4ca03081f7673f2) and a collection of roms, version files and softpatches as usual.


i30817:
None of them are unmanageable because of the script. Which is necessary because most of them update sooner or later. In fact, i expect next week 2-3 of the ones still in churn will have updated, maybe even on this one.

i30817:
Replaced a Phantasy Star III hack by a more complete translation that just came out today ;)

PR #622, opened 2018-04-22
add game boy hacks

i30817:
I improved my tool and made [a real project page for it](https://github.com/i30817/rhdndat), and now you can install it from pypi if you want.


RobLoach:
Although it doesn't work with ALL cores, there is [Softpatching ROMs](https://docs.libretro.com/guides/softpatching/) too.

PR #624, opened 2018-04-22
add game boy color hacks

i30817:
One of these will probably change when the description in romhacking.net changes, it's "Force GBA Enhanced Mode [Hack by Polar-Star]" which is a Shantae hack but the hack name doesn't actually reference the game. Thankfully users can propose changes on romhacking.net

edit: fixed.


RobLoach:
I suggest you look into softpatching too https://docs.libretro.com/guides/softpatching/ .... Don't really need these entries, but they're handy to have.

PR #625, opened 2018-04-22
add nintendo ds hacks

i30817:


RobLoach:
I could be wrong, but I think Softpatching works with NDS too. May need a docs update: https://docs.libretro.com/guides/softpatching/

i30817:
Last time i checked, it didn't. Anyway, even for platforms where softpatching works, when a VFS for softpatching happens in RA, these will be useful.

PR #631, opened 2018-04-23
add playstation hacks

i30817:
All the ps1 hacks i have. There aren't so many, but there is some gotchas here.

First, 3 translations here had to be modified by [error_recalc](https://www.romhacking.net/utilities/1264/) to work on the mednafen core. This is because the upstream has a 'bad' translation that nukes the error correction codes and mednafen objects. I thought it more 'reasonable' to include the working checksums not the broken ones.

The translations are Brigandine - Grand Edition, iS - Internal Section and Fantastic Night Dreams - Cotton Original.

Then there is the first game translation entry for front mission 2. This is a 'better' working version of the normal one on www.romhacking.net which actually has sound on videos among other things. Since i could never make the other one work without downloading the whole iso, and it was buggier and this one works with the redump iso, i thought i'd include this one instead. It has no romhacking page and the Mega link in which it was distributed is gone though, so it has to be maintained manually, unless i upload it to romhacking.net and it is accepted.

After this only the psp and ps2 games hacks are left for me. PSP i'm waiting for romhacking to review and accept two translations (one not so important `Last Ranker` and another quite important `Valkyria Chronicles 3`). For ps2 i know you don't have a core (yet) but it's only 6 of them.

### DIFF
--- a/metadat/hacks/Sony - PlayStation.dat
+++ b/metadat/hacks/Sony - PlayStation.dat
@@ -1,0 +1,151 @@
+clrmamepro (
+	name "PlayStation hacks"
+	description "Romhacks of Sony PlayStation games"
+)
+
+game(
+    name "Front Mission 2 (Japan) (v1.1) [T-En by pedr0]"
+    description "English translation by pedr0 version (1.1)"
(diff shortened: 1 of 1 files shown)


RobLoach:
I believe PlayStation uses serial scanning, so the CRCs may be completely ignored... Going to merge this though, so we can test further.

PR #652, opened 2018-06-10
Update GoodTools naming

RobLoach:
This fixing some of the GoodTools naming to match No-Intro's for N64 and WonderSwan/Color.

### Before
```
Dr. Mario 64 (U) [!]
```

### After
```
Dr. Mario 64 (USA)
```

Done programatically over at https://github.com/RobLoach/libretro-database-goodtools/pull/2 .


retro-wertz:
it would have been better IMHO that goodtools sets should be sorted and only include the info needed (im sure goodtools are added because of translations mostly), this should avoid most repeats and bad/overdumped roms unnecessarily increasing resulting rdb.

PR #660, opened 2018-06-26
Create Mattel - Intellivision.dat

RetroUp:
New No-Intro Intellivision Dat


RobLoach:
Thanks! Updated it in https://github.com/robloach/libretro-dats.

PR #661, opened 2018-06-26
Update Mattel - Intellivision.dat

RetroUp:
Updated from new No-Intro dat


RobLoach:
Thanks! Updated it in https://github.com/robloach/libretro-dats.

PR #665, opened 2018-07-04
new snes hacks

i30817:
Last time this was rejected because the database was getting too big, but i'm resubmitting in case you changed your mind and documentation for the future even if not accepted, separate PRs for nes and snes so they can be updated apart with force push.


RobLoach:
The majority of these already exist in:
https://github.com/RobLoach/libretro-dats/tree/master/static/Nintendo%20-%20Super%20Nintendo%20Entertainment%20System

The above is what build into:
https://github.com/libretro/libretro-database/blob/master/metadat/libretro-dats/Nintendo%20-%20Super%20Nintendo%20Entertainment%20System.dat

Thanks a lot!

PR #666, opened 2018-07-04
new nes hacks

i30817:
Last time this was rejected because the database was getting too big, but i'm resubmitting in case you changed your mind and documentation for the future even if not accepted, separate PRs for nes and snes so they can be updated apart with force push.

### DIFF
--- a/metadat/hacks/Nintendo - Nintendo Entertainment System.dat
+++ b/metadat/hacks/Nintendo - Nintendo Entertainment System.dat
@@ -27,3 +27,1912 @@ game (
     description "'Somari The Adventurer' hack by the jabu, Ti and Sics"
     rom ( name "Sonic_NES_Improvement_plus_Music_plus_Graphics.nes" size 786448 crc a38eceeb md5 0e2214993648014f95544b310ee6a11d sha1 b8a13b58d52d13fe1e616bb1c4fff6f827dc5890 )
 )
+game (
+    name "Cosmo Police Galivan (Japan) [T-En by Jair]"
+    description "English translation by Jair version (1.00)"
+    rom ( name "Cosmo Police Galivan (Japan).nes" size 262160 crc 4f031cb0 md5 08ecfd17a791acde0bf63543c98115a6 sha1 fd3d9f8a6c862773194c3e23b420aabe9453d2d6 )
+    comment "http://www.romhacking.net/translations/109/"
+)
+game (
+    name "Kamen no Ninja - Hanamaru (Japan) [T-En by pacnsacdave]"
+    description "English translation by pacnsacdave version (1.0)"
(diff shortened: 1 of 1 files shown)


RobLoach:
[Softpatching ROMs](https://docs.libretro.com/guides/softpatching/) may be an option here. You just stick the file right beside the original ROM, and it'll patch it for you when loading. Going to close this one for now, but let's keep our eyes open for interest for this to grow.

Thanks a lot.

OctopusButtons:
I agree there should be a database that maps *translation hack*  ROMs to the uniform game name (i.e. the game name recognized by RetroArch and which relates to thumbnails).  Ideally it would be in github so everyone can contribute, though I don't know how official members would verify the accuracy of each submitted hash = the stated game/hack.

Everyone agrees with something like this: _"Of course No Intro should have a database of hashes/checksums for verified relevant files, and names.  Of course our library-like apps (like RetroArch) should use the No Intro database to know what file is what.  Of course people want the machine to always know what file = what game name = what hash, and to get a thumbnail."_

I think almost any basis behind that widely-held consensus would also support the idea of doing the same for translation hacks (and other hacks beyond that).  Especially if the idea is a **retro library.**  We shouldn't limit ourselves to the mindset of "dumps of physical products" that No Intro is.

i30817:
It's a bit of a quixotic thing. First, RetroArch is always moving in the direction of magic strings\serials to identify roms, which is unsuitable to identify any hack, because they almost never change that.

Any person that seriously cares about hacks uses the manual scanner not to have missing hardpatched roms or duplicate softpatched roms. And the images are different, unless it's a translation and you keep the original name, but frankly there are no covers for the hacks but the originals, because again, the infrastructure is not there, mostly RetroArch would have to create the checksums itself because romahacking.net doesn't have a .dat and because RetroArch itself won't identify the hacks habitually while the automatic scanner is the default\exists.

RobLoach:
The problem is that hack IDs and names will always change, and there will always be an infinite number of hacks that we could add.

There is always the Manual Scan if needed, and we aren't opposed to adding the most popular few hacks. Just don't want to add unused bloat to the database.

i30817:
Two things are important to mention if someone in libretro-thumbnails goes the way of 'inventing' standardized hack names for images like I did in this PR.

RetroArch now has a ability to chop off metadata from the playlist name when looking for a image. This means a hack name, as long as it has a cutoff name in the thumbnails server can have a single image if metadata that varies a lot is inside brackets (like the version number).

I did it another way in this PR, if you'll notice I put in the version number and details of any extra combinatoric hack in the description field, which is also a possibility. I did it because putting all the details would often cause huge labels that have to scroll left and right, but it also works to uniquify a thumbnail if those details include the version.

PR #701, opened 2018-09-09
Sanitize Spectrum cheats

RobLoach:
@leiradel Unsure how RetroArch .cht format reacts to double quotes within double quotes. Lets just remove them.


leiradel:
Good catch, I think this is safe to merge.

PR #718, opened 2018-10-07
Add more Legend of Zelda 1 cheats.

wrldwzrd89:
Adding various cheats for The Legend of Zelda (NES) from GameHacking.org. Tested in Nestopia and verified to work.

### DIFF
--- a/cht/Nintendo - Nintendo Entertainment System/Legend of Zelda, The (USA, Europe) (Action Replay).cht
+++ b/cht/Nintendo - Nintendo Entertainment System/Legend of Zelda, The (USA, Europe) (Action Replay).cht
@@ -1,4 +1,4 @@
-cheats = 37 
+cheats = 36 
 
 cheat0_desc = "Lessens Damage"
 cheat0_code = "000670FF"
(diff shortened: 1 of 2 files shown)


cht/Nintendo - Nintendo Entertainment System/Legend of Zelda, The (USA, Europe) (Action Replay).cht
```diff
@@ -1,4 +1,4 @@
-cheats = 2 
+cheats = 37
```
RobLoach: Believe this should be 36?

RobLoach:
Nice work! Happy to have these in place. Mind updating the total?

cht/Nintendo - Nintendo Entertainment System/Legend of Zelda, The (USA, Europe) (Action Replay).cht
wrldwzrd89: There are in fact 37 entries, but one of them has no code. That's because it's intended as a note - the Merchant Sells cheats can't be used together. It sounds like the format doesn't support notes, though. Super-easy to fix.

wrldwzrd89:
Just updated to get rid of the note and revised the cheat total to 36. Renumbered all cheats after the removed one.

RobLoach:
:+1:

PR #728, opened 2018-10-16
changed 'comment' tag for 'patch' on the metadata/hacks dir

i30817:
Skipped actual comments.

This is to make the RA frontend be able to treat 'patch' (which contains a url to the page where the patch can be downloaded and 99% of the time has a description - always on romhacking.net urls) especially and be able to load on the system browser all 'patches' if the user wants more information on what the game is.

Previously, 'comment' could be anything and often was. Logs from cowering tools, hints that a hack has special requirements etc. This way a special button in the info page can be created to open all 'patch pages' on the browser, which i'm going to ask for right after this is merged.


i30817:
Some thoughts:
a 'game' can have more than one 'patch' (ie: duplicated tags at the same level), is this ok for retroarch and libretro database?

PR #732, opened 2018-10-18
Further fixes to "Pokemon - Emerald Version (USA, Europe) (Code Breaker).cht"

autofire372:
Fix a typo, and re-index the cheat list so it starts from 0, not 1.


autofire372:
Turns out the GBA cores need the encryption constant to be present in every code group, not just the master code.

PR #741, opened 2018-11-13
Add MAME 2003-Plus (2018-11-11) DAT

markwkidd:
I would like to convert this all the way to RDB format but the README docs don't really tell me how to go from DAT to RDB for one file: https://github.com/libretro/libretro-database#building


RobLoach:
Mind renaming this to `MAME 2003-Plus.dat` so that it matches in the DB? Thanks!

markwkidd:
1. No problem, done. The date serves as the version of MAME 2003-Plus but it's in the header of the file itself.
2. Yep these are all torrentzipped

PR #744, opened 2018-11-15
correct readme instructions for libretro-build-database script

markwkidd:


markwkidd:
Rather than correcting the copied the build instructions from libretro-db/README.md here, I am instead going to suggest linking to the "canonical instructions" and fixing them. That concept is reflected in the latest commit to this PR.

PR #745, opened 2018-11-19
snes: Add Super Famicon Wars (Japan)

RobLoach:
@meepingsnesroms was looking to play this netplay. While soft-patching workds, having it in the menu makes life easier.

### DIFF
--- a/metadat/hacks/Nintendo - Super Nintendo Entertainment System.dat
+++ b/metadat/hacks/Nintendo - Super Nintendo Entertainment System.dat
@@ -36,5 +36,5 @@ game (
     name  "Super Famicon Wars (Japan) (E)"
     description "Super Famicon Wars (Japan) (E)"
     homepage "https://www.romhacking.net/translations/3354/"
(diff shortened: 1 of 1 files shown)


i30817:
Ok, _now_ can i add the fairly complete dat list of nes and snes hacks you rejected before?

This was there - among with many other missing not not updated patches.

RobLoach:
> Ok, now can i add the fairly complete dat list of nes and snes hacks you rejected before?

All? No... But the most popular ones, yes.

i30817:
Yeah screw that. I prefer to figure out how to create my own rdb if you're going to keep sabotaging this one to 'parcials'. Though admittedly it's kind of useless atm because retroarch upstream is sabotaging itself with serials.

i30817:
I seriously don't even understand the point of complaining about performance of the _database_ of all things. I'm sure that you're aware that the scanner dominates (crushes) the performance cost of that usecase compared to the lookup and that cost is directly proportional to the size of the rom, not the number of entries on that database.

RobLoach:
One way to show this would be some benchmarks. A concern that Twin had was for low-powered devices. It would be possible to spin up a libretro-database-extras with the entire collection.

Just don't want to add so many since we're not truly optimized and battle-tested database system in RA.

PR #747, opened 2018-11-27
update mame2003-plus DAT and RDB

markwkidd:
This adds a few newly supported mame2003-plus romsets to the DAT and RDB. Notably Virtua Fighter!

This PR also benefits from a bugfix to my romset sorting script which was mistakenly including a number of games with non-working romsets that I originally meant to exclude. That is why there are lines being removed.


RobLoach:
Thanks for keeping these up to date!

PR #752, opened 2018-12-12
update mame2003-plus database

markwkidd:
Perhaps most importantly, this update will mean that there is 100% coverage for mame2003-plus Snapshots and Title Screen thumbnails as demonstrated over here: https://github.com/markwkidd/mame2003-plus-thumbnail-sources

There has been a round of work on the early Namco drivers which changed some romsets; the changes are reflected here.

Also I have, I think, fine tuned my script for only including working mame2003-plus romsets in this database, so there there are some more non-working romset entries removed that should have been removed before now.


markwkidd:
I fixed the merge conflict

RobLoach:
Thanks! Happy that you've taken this on.

PR #754, opened 2018-12-22
Add cheat to Kirby Super Star (USA) (Game Genie).cht

robertxgray:
Cheat added for Kirby Super Star (snes). Same as 62E2-AD61 but for player 2.


RobLoach:
Thank you

PR #788, opened 2019-01-23
update MAME 2015 database

markwkidd:
As with my other recent PRs, this one will improve the accuracy of the MAME 2015 db. It's based on having both a more correct (MAME 0.160 vs 0.159) DAT but also an expanded DAT with more and better metadata.

That expanded metadata is the reason why the MAME 2015 DAT must now be in a ZIP file -- in its uncompressed state, that file is larger than the 100MB github filesize limit.


markwkidd:
Thanks! Till lately we've been saying use MAME 0.159 romsets when it really should have been 0.160. 

This core has enough issues without romset mismatches 😋 hopefully this will help people have a better experience

PR #792, opened 2019-01-27
Add data for Thomson MO/TO computers (theodore core)

Zlika:


RobLoach:
Looks good! Anything needing updates before bringing it in? Thanks a lot for pushing this forwards.

Zlika:
Thanks.
What do you have in mind by "other changes that are needed too"? I'm still discovering how are working the libretro DB, assets, thumbnails..., there is not lot of documentation about it.

PR #799, opened 2019-02-06
add MAME 2016 DATs to RDB database

markwkidd:
This PR adds Full Non-Merged and Split DATs for MAME 2016

@RobLoach do these files need to be added to any scripts, or will the RDB generation process automatically find these?


RobLoach:
Just in that libeetro-build-database script in libeetro-build-database! Nice work 👍

markwkidd:
Thanks -- looks like there is already a rule in place! https://github.com/libretro/libretro-super/blob/master/libretro-build-database.sh#L332

PR #803, opened 2019-02-17
Added Sega Genesis Hacks

Europia79:
- Sorted Alphabetically
- Added:
NHL '94
Psy-O-Blade
Surging Aura

### DIFF
--- a/metadat/hacks/Sega - Mega Drive - Genesis.dat
+++ b/metadat/hacks/Sega - Mega Drive - Genesis.dat
@@ -3,51 +3,10 @@ clrmamepro (
     description "Romhacks of Sega Mega Drive / Sega Genesis games"
 )
 game (
-    name "Rock n' Roll Racing (USA) Hack"
-    description "Hack by Ti version 15"
-    rom ( name "Rock_n'_Roll_Racing_Hack_v15.bin" size 2097152 crc 205f4afc md5 bd4bae13a6fed5d2801face18572fab2 sha1 bedfa111535fbeea0b75999b799317415bc4bab3 )
-)
-game (
-    name "El Viento Enhancement [Hack by M.I.J.E.T.]"
-    description "El Viento Enhancement by M.I.J.E.T. version (101010)"
-    rom ( name "El. Viento (USA).md" size 1048576 crc 363946b2 md5 ca18abf6ccede48681f013db3c43f108 sha1 83c55f390ce33cfbfcd8c913e8a34a0c3bc304ae )
-    patch "http://www.romhacking.net/hacks/678/"
-)
-game (
-    name "Rent a Hero (Japan) [T-En by NikcDC]"
-    description "English translation by NikcDC version (0.97)"
-    rom ( name "Rent a Hero (Japan).md" size 1271970 crc efe3483b md5 ce0affb925c93c4ad8342ac874af87cd sha1 0b2f5da12190cd9ecf1dfd374781b2ff3931635d )
-    patch "http://www.romhacking.net/translations/2309/"
-)
-game (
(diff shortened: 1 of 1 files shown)


RobLoach:
Thank you!

i30817:
Does it matter they're sorted? Because i made the previous versions on purpose to clearly divide those hacks that were 'not done by my tool' (rhdndat) from those that were manually added.

The advantage of using rhdndat is that i don't have to care about manual updates and the diffs stay minimal and easy to review. So i'm basically asking this is reverted to place the new entries on top unless it's really important.

i30817:
Also @RobLoach, is there a comment marker that works for ignoring a line in these dat files? I'm thinking of adding a line `#following entries added by rhdndat` to make the division explicit.

Europia79:
So... I wanted to add some entries... So naturally, the question arises, "where do I put them ?"  I mean, there's several different strategies.  I could prepend new entries at the start of the file ?  I could append new entries to the end of the file ?  I could insert new entries in the middle of the file ?  But with the lack of an apparent standard, it's not immediately obvious to new-comers (like me) what the correct answer is.

And a logical argument can be made for all three strategies.  But I disagree that there should be different rules for different people.  There should be one rule and one standard for all people to follow... including automated tools.  This is just the nature of software development:  That tools are made to specification.

So really, we both have the exact same problem:  No standardize rule for writing new entries.

I mean, apparently, it's a problem for your tool ?  Altho, I don't exactly understand why ?  Can you elaborate on why your tool needs a division between it's own automated entries and the entries generated by other tools ?  What exactly happens when there's no division and they're all mixed together ?

Europia79:
Feel free to call me on Discord:  I'm `Europia79#7146`

I'm also in the Discord servers for `romhackingdotnet` and `libretro`

I hate unresolved issues like this, and I feel like it'd much faster to collaborate over a quick discord session, knock it out, then come back & post a solution that everyone will be happy with.

Also, thanks for pointing out those errors.

RobLoach:
> Does it matter they're sorted?

No. Order doesn't matter.

> The advantage of using rhdndat is that i don't have to care about manual updates and the diffs stay minimal and easy to review. So i'm basically asking this is reverted to place the new entries on top and ditch the sorting unless it's really important, and if so i can sort the rhdndat entries myself too.

That's okay, we can review in the diff.

> Also i'd suggest not naming hack roms with (USA) in the name. The standard which my tool has been following is the original rom name for translations followed by [T-Lang by authors] and the hack name for hacks followed by [Hack by authors] and consistency is user friendly.

`(USA)` is the region. It's helpful to have that to match the source rom region. You can have both (USA) and [T-Lang] flags.

> is there a comment marker that works for ignoring a line in these dat files in the tools that process them?

You can have `comment` entries, similar to:
https://github.com/libretro/libretro-database/blob/master/dat/BIOS.dat#L15

> I hate unresolved issues like this, and I feel like it'd much faster to collaborate over a quick discord session, knock it out, then come back & post a solution that everyone will be happy with.

There's no unresolved issue here. We've done our best to place DAT files in logical places. The documentation and formatting could improve, but this is working for now. If you have any suggestions, feel free to create a new issue. :+1:

i30817:
> I mean, apparently, it's a problem for your tool ? 
No, it works fine. Sort of, it's complicated code to match, but at most you'll end up with duplicate entries you'll need to correct. 

The problem such as it is, is that for me it's much easier to check what few romhacks are not automatically checked if they start on the start of the file, and to remove duplicates in romhacking.net gets it and i can use the tool to check for updates.

Europia79:
I"ll fix that mistake now... From looking at the RetroArch database, it appears that it uses the fully-qualified names for the `name` & `description` fields and the abbreviation for the `rom_name` field.

Issue #635, opened 2018-05-03
v0.2.97.43 Arcade FBA new games are not listed in retroarch playlist


furiadeoso:
@RobLoach Hi! Could you update FBA core dat files here? A new version was released.
I would do it myself if I knew. The XML of my romset has a very different format from the one used here.
Thanks.
Best Regards!

RobLoach:
@bkoropoff

retro-wertz:
dat files are still not updated to 0.2.97.43

barbudreadmon:
https://github.com/libretro/fbalpha/issues/208
https://github.com/libretro/fbalpha/issues/200

ParzivalWolfram:
You should move this to the appropriate core's repo or the main RetroArch repo.

furiadeoso:
There is a new FBA version, and database for playlist building is very outdated (since 0.2.97.42). Here is the new DAT: https://github.com/libretro/fbalpha/blob/master/dats/FB%20Alpha%20v0.2.97.44%20(ClrMame%20Pro%20XML).dat

Without updating is almost impossible to use retroarch without an external frontend (or at least very uncomfortable).

Who is aware of how it is updated?

By the way, the thumbnails are also quite outdated, where is your repository?
Thank you.

barbudreadmon:
As i have already said a long time ago, those database thingy should ignore checksums and only rely on filenames for arcade systems, it would make your lifes SO much easier. Hopefully i'm using an external frontend personally.

furiadeoso:
If this is not solved it will not work well the new ask-to-join discord integration with fba. And netplay is practically dependent on an external frontend.

retro-wertz:
> 
> 
> As i have already said a long time ago, those database thingy should ignore checksums and only rely on filenames for arcade systems, it would make your lifes SO much easier. Hopefully i'm using an external frontend personally.

i agree with this (or the concept anyways) but this would mean a special "scan arcade romset" should be added? is it possible for current scanning to be manipulated to search match of "zip filename" instead of the crc? @RobLoach

ParzivalWolfram:
I wasn't thinking it'd be blocking, more of a "these codes might not work,
you sure?" type of thing.

PR #804, opened 2019-02-17
Update FBA

alfrix:
Split with merged samples 
torrentzipped
dat exported with https://github.com/libretro/libretro-database/blob/master/scripts/FBA_dat_gen.py
source dat: https://github.com/libretro/fbalpha/blob/master/dats/FB%20Alpha%20(ClrMame%20Pro%20XML%2C%20Arcade%20only).dat

This matches the current core and thumbnails

https://github.com/libretro/libretro-database/issues/635


RobLoach:
WOOO! :+1:

PR #816, opened 2019-03-09
Unique spelling for Squaresoft in developer and publisher

kivutar:
It was sometimes spelled `SquareSoft` resulting on wrong having two different developer pages for the same trademark.


kivutar:
It's SquareSoft

PR #858, opened 2019-05-22
Add a few Super Metroid Rom Hacks to SNES.dat

frozenLake:
Hacks in question added were Hyper Metroid, Project Base, and Ancient Chozo, with Homepages pointing at Metroid Construction.
Probably could do with having more hacks be added to the hacks section in general, but I figure this might be a good start.


RobLoach:
Since [Soft-Patching in RetroArch](https://docs.libretro.com/guides/softpatching/) works, and is an easier method to get the hacks applied, we opt to keep only the most popular ones in here. Adding all of the hacks bloats the database, so we'd like to keep it slim when possible.

These look like good ones though! Thanks a lot.

PR #866, opened 2019-06-02
updates for new and old hacks

i30817:
Some updates.

A crc changed in the same version but it was just because it was written in uppercase before.


i30817:
Had to force push a name correction because i left around a rom that shouldn't have been there.

PR #870, opened 2019-06-03
Update Sinclair ZX81 DAT

mnadareski:
This makes the DAT one that is recognized by most of the prevailing programs out there for management.
Notable changes:
- ~Removed duplicate set names by combining ones with the same name~
- Removed instances of `:` and `/` from set names (reserved path characters that can't be used)
- ~Automatic alphabetization of the DAT~
- Fixed duplicate set names by appending file type

### DIFF
--- a/dat/Sinclair - ZX 81.dat
+++ b/dat/Sinclair - ZX 81.dat
@@ -1,6089 +1,3149 @@
 clrmamepro (
-  name "Sinclair - ZX 81"
-  description "Sinclair - ZX 81"
-  version 20160505113721
-  comment "extracted from http://www.zx81stuff.org.uk, also with games from http://www.bobs-stuff.co.uk/zx81.html"
+	name "Sinclair - ZX 81"
+	description "Sinclair - ZX 81"
+	version "20160505113721"
+	author ""
+	comment "extracted from http://www.zx81stuff.org.uk, also with games from http://www.bobs-stuff.co.uk/zx81.html"
 )
-
 game (
-  name "10 Games (J. K. Greye)"
-  description "10 Games (J. K. Greye)"
-  year "1981"
-  manufacturer "J. K. Greye"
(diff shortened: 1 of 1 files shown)


dat/Sinclair - ZX 81.dat
```diff
@@ -1,6089 +1,3149 @@
...
-  name "Fast Load Save (Musamy Software)"
-  description "Fast Load Save (Musamy Software)"
-  year "1982"
-  manufacturer "Musamy Software"
-  rom ( name "FastLoadSave.tzx" size 11000 crc 939F03CB md5 BE316FADC3492E475E6D07BDF9AE310E sha1 616DFC2662D46F415277990BEBF0B09E187ED36E )
-)
-
-game (
-  name "The Fast One (Campbell Systems)"
-  description "The Fast One (Campbell
```
RobLoach: I see multiple `rom` entries. in some of the `games`. RetrArch playlists don't really work with multiple entries. I'm honestly unsure how it reacts to it.... Does the core care if it picks up the .p or the .tzx?

mnadareski: That I'm not sure of. I just fixed this according to what DAT managers would normally work with. The other option would be to do something like postfixing the game names with `[P]` or `[TZX]` to determine what one it is. That's not the worst option, though it does mean a lot more manual labor to fix the file.

mnadareski: At the very least, I can update this to only fix the `:` issue for now. Most DAT managers can cope with duplicate set names.

mnadareski: I found a reasonable way to deal with this, see the latest revision.

RobLoach:
Hmmm, I'm not sure about splitting them up entirely and having users select. Could we just select the one that launches well in the core?

The other things we could do is just roll with what you had before, with both files. Build it, and let people test it out a bit.

mnadareski:
Either option is fine by me, I mainly just wanted to make sure that this DAT was up to standards. The rest of this (the deduplication of game names, etc) is just extra fluff.

RobLoach:
Great! I'll merge this, and then revert the split commit. Feel free to test away!

PR #881, opened 2019-06-20
Update MegaDrive\Genesis BIOS

LodanZark:
Recently was listed in the no-intro database a clean MD BIOS file, I already test it with Genesis GX Plus and seems to work fine.
Source: [https://datomatic.no-intro.org/index.php?page=show_record&s=32&n=0007](url)

To make the life easier I also created a BPS patch to convert the bad dump into a good dump that can be found here: 
[Fix MD BIOS.zip](https://github.com/libretro/libretro-database/files/3312387/Fix.MD.BIOS.zip)


RobLoach:
giphy

PR #887, opened 2019-06-25
make databases buildable without libretro-super

fennectech:
running libretro-build-databases will regenerate the databases without the need for the libretto-super being needed making it far less cumbersome for those not looking to deal with reproach.


fennectech:
it's not a perfect solution but it's far cleaner than dealing with libretro-super if you only want to build the databases.

PR #908, opened 2019-07-26
added Yume Nikki and Yume 2kki on RPG Maker dat

maxlefou:
title ☝️ 

Couldn't add sizes, it doesn't seems to be in bytes or anything. How you add them? (especially if you just have windows)


RobLoach:
Looking great! The size is in bytes, yes.

To get the size of the file it should be `dir *.ini`...... or `ls -al *.ini`

maxlefou:
Thanks. My main problem was actually me having the size in octets and not in bytes.. But i could convert the numbers and get it right. They're added now.

PR #927, opened 2019-09-18
Create Resident Evil 3 - Nemesis (NTSC-U) (USA, Europe) (GameShark).cht

Gillabong:
Working codes for use with retropie


RobLoach:
Thanks! :+1:

PR #929, opened 2019-09-19
Create Comix Zone (USA) (GameGenie)

Gillabong:
another working game genie add for sega megadrive-genesis on retropie. I have found that action replay cheats do not typically work well or at all on retropie for sega megadrive-genesis


RobLoach:
This game was so unique. Loved it.

PR #933, opened 2019-09-19
Create RoboCop versus The Terminator (USA) (GameGenie)

Gillabong:
wanted to add this list of working codes since the action replay list currently available for this game does not work properly


RobLoach:
Thanks so much for putting some attention into the cheats. They've been neglected as of late.

Gillabong:
No problem. If it's ok with you, I would like to add more corrections.

RobLoach:
Most definitely! Shoot any PRs and we'll make sure they get merged.

PR #942, opened 2019-10-10
vitaquake2: Add Quake II Database

RobLoach:
This adds a Quake II database so that Quake II can be scanned and loaded through the menu.

@Rinnegatamante Mind reviewing these pak0.pak files? Unsure if they're the right files, but they seem to be loading correctly to me.

### DIFF
--- a/dat/Quake II.dat
+++ b/dat/Quake II.dat
@@ -1,0 +1,34 @@
+clrmamepro (
+	name "Quake II"
+	description "Quake II"
+	version 2019.10.10
+	author "libretro"
+)
+
+game (
+	name "Quake II"
+	description "Quake II"
+  homepage "https://en.wikipedia.org/wiki/Quake_II"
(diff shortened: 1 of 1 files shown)


Rinnegatamante:
Demo files are included in VitaDB download (https://vitadb.rinnegatamante.it ). Full files can vary depending on where they come from (GOG and Steam versions are slightly different i think). Btw they look fine.

RobLoach:
> Demo files are included in VitaDB download

Great! Looks like that pak0.pak's CRC matches what's in the database here. Thanks. 



> GOG and Steam versions are slightly different i think

I got that one from Steam. Don't have the GOG version, unfortunately. Made an issue: https://github.com/libretro/libretro-database/issues/943

THANKS A LOT!!!! THIS IS THE BEST CORE IN THE WORLD!@!!@@!!!!!oneoneone

PR #953, opened 2019-11-13
Localization support

guoyunhe:
This is an attempt to add localization support of game names.

*.dat files --> generate template *.ini files --> sync with crowdin (github integration)

Translations will in `l10n` directory. Each sub-directory will be a language (when pulled from Crowdin).

Translations will be a new option in Online Updater: users can download only the language they need, to save space and bandwidth. For English speaking users, they don't need these files at all. This is why we don't compile it to dat or rdb files. Translations are used to generate localized play lists.


RobLoach:
Great work, looks like a neat solution around it... How will this work with the RDB files?

guoyunhe:
These files are provided along with RDB files but they are separated. Translation files should be optional.

RetroArch, or other clients, need to download both of them.

It first searches hash in RDB and get the English name, e.g. "Yu Yu Hakusho - Makyou Toitsusen (Japan)". Then it matches the name in INI file and get the translated name "幽☆遊☆白書 魔強統一戦 (日本)" (according to users' language preference).

schellingb:
I don't think this approach is meaningful in this form. Besides the issues of using the name (that can change in the source at any time) instead of a checksum or a serial number to connect entries, I don't think game names are a translation issue.

RDBs contain one entry per ROM and not one entry per game. So for most systems, a ROM shows a title screen in just one language.
In my opinion whatever that ROM shows on the title screen should be stored in the 'name' field of the RDB entry. Then there could be an additional field 'ascii_name' or 'romanized_name' which contains a (English) translated name (if common) or a transliteration of the name using just basic latin letters (if different from 'name').

Now for systems like NeoGeo where a single ROM often (?) covers multiple regions with different titles there needs some additional thought put behind. But I think some of these games already show different English titles depending on PAL and US region so it isn't really a new problem. For those I think that 'name' should be the "most original" name (let's say the name the original developer would most likely use to refer to that game). Then there could be an additional field in the database "alternate_names" which contain all other names. Or there could be 'name_jp', 'name_kr', 'name_cn' and so on (maybe 'name_pal'?').

RobLoach (changes requested):
While I think localization would be awesome, I think we should consider how we take this on. Having the support for it in RetroArch first would be nice.

PR #955, opened 2019-11-13
Cave Story (World) (v0.5.2) - ReleaseYear

amirrammari:


RobLoach:
May be best to have them all in the same entry:
https://github.com/libretro/libretro-database/blob/master/dat/Sega%20-%20Mega%20Drive%20-%20Genesis.dat#L429-L433

Mind moving this there? Thanks!

Issue #925, opened 2019-08-31
FBNeo importing content only works when Scan Without Core Match: ON


RobLoach:
From @rsn8887  

## Description

When scanning any FBNEo rom via `Import Content`, e.g. my crc32-verified rom 1941.zip, it doesn't show up in the FBNeo playlist. I have the core installed, everything updated etc.

When I turn ON `Settings->Playlists->Scan without core match`, then it works.

### Expected behavior

The rom should appear in the FBNeo playlist also with `Settings->Playlists->Scan without core match` OFF as long as I have the core installed.

### Actual behavior

See above

### Steps to reproduce the bug

1. Import Content
2. Import Single File
3. Select a FBNeo rom, e.g. 1941.zip
4. Observe it is not added to playlist
5. Repeat after Settings->Playlists->Scan without core match:ON
6. Observe working

### Bisect Results

Never worked.

### Version/Commit
You can find this information under Information/System Information

- RetroArch: Latest Nightly 08/30/2019

### Environment information

- OS: Switch

RobLoach:
@rsn8887 .... What's the CRC of the file? What playlist does the file get added to? More info would be helpful. Thanks!

rsn8887:
The file I tested was 1941.zip and I checked all crc32 matched the fbneo dat file entry from the current libretro-databases repo here.

I think my issue might be the same issue as https://github.com/libretro/libretro-super/issues/1107

According to @metchebe the fix is to remove the line database_match_archive_member = "true" from the core info file.

rsn8887:
After setting Settings->Playlists->Scan without core match:ON and re-scanning, all the games appear. However, when I now go to Online Updater->Playlist Thumbnail Updater, the playlist is still listed as "FB Alpha - Arcade Games", not "FB Neo - Arcade Games"

metchebe:
@rsn8887 do you still have the old FB Alpha rdb in the database folder? It should be deleted. My playlist shows up correctly as FB Neo.

rsn8887:
You are right I still had the old database. I also have the new FBNeo database.

I deleted the old FB Alpha database and Playlist. When I now scan my FBNeo rom folder, no playlist appears even though the roms are correct.

So the only reason it worked before was because it was using the old database.

RobLoach:
I'm assuming you've done the same with the cores, and downloaded the Neo one too?

rsn8887:
[...]

PR #966, opened 2019-12-04
Update FBNeo

alfrix:


RobLoach:
Thanks! Submitted a PR to fix up the db build script.

metchebe:
I could be wrong, but I believe this update has made the FB Neo database incompatible with current valid un-merged torrentzipped roms. I believe it has to do with the torrentzip comment being removed from the reference .zip set.

Commented initially here: https://github.com/libretro/libretro-database/issues/925#issuecomment-569773513

PR #972, opened 2020-01-12
Update Sony - PlayStation.dat

Justme488:
http://redump.org/disc/672/


RobLoach:
Mind adding these into here instead?
https://github.com/RobLoach/libretro-database-psxdatacenter/blob/master/overrides.js

Like this.....

``` js

module.exports = function (title) {
	return title
		.replace('shell', 'Shell')
		.replace('wheelman', 'Wheelman')
		.replace('never dies', 'Never Dies')
		.replace('Saga Frontier', 'SaGa Frontier')
		.replace('Defcon 5 - Piece Has a Price', 'Defcon 5 - Peace Has a Price')
		.replace('Arc the Lad Collection (USA) (Disc 1 of 6)', 'Arc the Lad Collection: Arc the Lad (USA)')
}

PR #1006, opened 2020-02-26
Repaired Wolfenstein 3D.dat

Sanaki:
Changed file to rom in each entry to allow the rdb to build correctly.


Sanaki:
I can add the rdb I compiled from that if desired, but I assume for compiled files a more "official" source would be preferable.

RobLoach:
Thanks for the fix :+1:

PR #1018, opened 2020-03-02
Includes shareware wolf v1.1 and v1.2.  Includes retail wolf v1.1

pkos:
Updates for phcoder's release of the new ecwolf.pk3

### DIFF
--- a/dat/Wolfenstein 3D.dat
+++ b/dat/Wolfenstein 3D.dat
@@ -6,6 +6,48 @@ clrmamepro (
 )
 
 game (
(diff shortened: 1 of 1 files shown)


pkos:
Versions 1.4 and 1.4g shareware offer nothing different but the exe
so they will both be picked up by v1.4 (Shareware)
Versions 1.4 1.4c and 1.4g retail offer nothing different but the exe
so those three will be picked up by v1.4

RobLoach:
Thanks for the fixes!

PR #1036, opened 2020-04-01
Add HBMAME.dat

Zoltan45:
Add HBMAME.dat


RobLoach:
Thanks for the contribution! A few things....

```
machine (
	name 1942
	description "1942 (Revision B)"
	year 1984
	manufacturer "Capcom"
	rom ( name srb-03.m3 size 16384 crc d9dafcc3 sha1 a089a9bc55fb7d6d0ac53f91b258396d5d62677a )
	rom ( name srb-04.m4 size 16384 crc da0cf924 sha1 856fbb302c9a4ec7850a26ab23dab8467f79bba4 )
	rom ( name srb-05.m5 size 16384 crc d102911c sha1 35ba1d82bd901940f61d8619273463d02fc0a952 )
        ...
)
```

- RetroArch RDB builder uses one `rom` entry per game, so we won't need multiple in here. Make sure they're unique to the rom too. Alternatively, you could use a TorrentZip of the archive, similar to some of the other MAME DATs.
- Use `game` instead of `machine`
- `releaseyear` instead of `year`
- `publisher` or `developer` instead of `manufacturer`

Zoltan45:
ok. working on figuring out the best way to get similar datfiles right now.  If you have any suggestions I am all ears.

Thanks,

PR #1037, opened 2020-04-02
Add HBMAME datfile

Zoltan45:
add HBMAME datfile


RobLoach:
This looks great!!! Are these torrentzipped? Mind adding a `description` in the header describing how it was built?

Zoltan45:
The roms in the datfile are torrentzipped yes.  Do you mean a description of how the romset is built or the datfile? @RobLoach

RobLoach:
This is perfect.

PR #1043, opened 2020-04-12
Create Asterix and Obelix - Bash Them All! (Europe) (En,Fr,De,Es,It,N…

mixpc:
Codes taken from the Internet.
Note Asterix & Cleopatra Infinite Lives seems to block turning and hitting left.


RobLoach:
Thanks!

PR #1047, opened 2020-04-28
Clean up hacks (resolves #1040)

### LINKED ISSUE #1040: ROM Hack / Homebrew Naming Convention
Is there a proper naming convention for ROM hacks, homebrew, or anything else outside of the No‐Intro database? The dats here seem a bit inconsistent.

rouhannb:
I’ve applied some more consistent naming to the ROM hacks.

I’d also like to ask what should be done for translations. Some of them translate the title, but all the entries here use the original title. Should this be changed?


RobLoach:
Looks great! Thanks a lot.

PR #1060, opened 2020-05-14
Create Megaman Battle Network 4 : Red Sun

fdiaz93:


RobLoach:
Mind renaming it to `.cht`? Thanks!

PR #1090, opened 2020-09-14
Update 3DO firmwares to match with Opera libretro core

LodanZark:


LodanZark (commented):
3DO firmware update changes to reflect Opera libretro core ( https://docs.libretro.com/library/opera/ )

RobLoach:
Thank you :rocket:

PR #1101, opened 2020-10-14
CannonBall Complete Set Added

LodanZark:
Files list based on the official git:  https://github.com/djyt/cannonball/blob/master/roms/roms.txt

PS: epr-10381.132, epr-10381a.132 and epr-10381b.32 has the exact same hash, the explanation why here: 
https://github.com/djyt/cannonball/commit/680b5246bce377eccc9a6ffc27d29dd482caf708#commitcomment-43242378


RobLoach:
Thanks a lot for taking a look at this. Not sure about having multiple files listed. RetroArch and the database take one game -> one rom approach, which lets the menu know which file to display in the playlist.

LodanZark:
Yeah, makes all sense, if cannonball core would accept compressed files, would make things easier :S

RobLoach:
> if cannonball core would accept compressed files

This has been on my wishlist FOREVER, and here's a related issue for it: https://github.com/libretro/RetroArch/issues/4331

LodanZark:
Adding this filename and hashes to the core info would make the trick, but at the sametime it would lose the ability to curate the roms through rom managers like clrmamepro and other alternatives, I will let you decide what it's the best option ;)

RobLoach:
Just having one is optimal for the RetroArch database. Could keep the entire set elsewhere though! If we have multiple in there, I'm unsure how libretro-build-database.sh would respond. It might end up listing a bunch of the same entries in the playlist, which would seem silly.

This could live in a dat in the source repository or something?

PR #1106, opened 2020-10-28
More BIOS

LodanZark:
ADDED:
Fairchild - Channel F
NEC - PC-98
Sharp - X1


RobLoach:
You're on fire! :fire:

PR #1108, opened 2020-11-02
Update - Arcade BIOS (fbneo, flycast)

LodanZark:
Added BIOS needed to boot arcade games for flycast and fbneo core, also can be used for mame too.
Files are torrentzipped to keep hashes consistent.

### DIFF
--- a/dat/System.dat
+++ b/dat/System.dat
@@ -2,7 +2,7 @@ clrmamepro (
 	name "System"
 	description "System"
 	comment "System, firmware, or BIOS files used by libretro."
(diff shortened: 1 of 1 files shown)


LodanZark:
TorrentZip Tool: https://www.romvault.com/trrntzip/
MAME 0.226 BIOS DAT
[MAME 0.226 ROMs (bios-devices).zip](https://github.com/libretro/libretro-database/files/5476722/MAME.0.226.ROMs.bios-devices.zip)

PR #1116, opened 2020-12-28
Use `make install` on the Gitlab CI to set up

RobLoach:
Use the [Makefile](https://github.com/libretro/libretro-database/blob/master/Makefile) to set up libretro-database. Before the chts dir was missing. Using `make install` will include it.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -7,11 +7,7 @@ stages:
   variables:
     MEDIA_PATH: .media
   script:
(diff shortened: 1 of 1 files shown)


RobLoach:
@jdgleaver Haha. Looks like you've fixed it :stuck_out_tongue_winking_eye:

jdgleaver:
Ah, sorry! I didn't notice your PR!

Using the Makefile would perhaps be better than my manual copy - so do feel free to make this change, if you like :)

PR #1128, opened 2021-02-04
adding info on featured LowResNX games

afauroux:


fpscan:
What do you think if we change the names to more basic level, like I did here: https://github.com/libretro/libretro-content/pull/19 so when file updated there will be no need to change dats, thumbnails etc.

PR #1133, opened 2021-02-21
Added nkit.gcz (compressed) versions into GameCube dat

adhrymes:


adhrymes:
Can I add multiple 'name' entries in game (..) block? If so I can update to do it like this

RobLoach:
Not sure this accomplishes what you think it does. The file name is pretty much ignored when using serial scanning.

adhrymes:
> 
> 
> Not sure this accomplishes what you think it does. The file name is pretty much ignored when using serial scanning.

Hmm how can i have it recognize the nkit.gz files then?

RobLoach:
RetroArch won't automagically recognize large compressed files like that. You'll have to use the manual scanner for that.

adhrymes:
> 
> 
> RetroArch won't automagically recognize large compressed files like that. You'll have to use the manual scanner for that.

ahh gotcha, I was able to do this and have them added to my playlist properly but the thumbnails arent showing up properly for the compressed versions. What should I update for them to be read in properly in bulk (as opposed to manually editing the playlist file)

RobLoach:
Thumbnail filenames must match the name of the entry in the playlist.

PR #1135, opened 2021-03-02
Create Parasite Eve (World) (GameShark).cht

IceLancerSR:
It includes two retroarch memory code for infinite tools. Tested.


RobLoach:
This is great! Didn't know you could mix the memory with the gameshark codes. Nice work!

PR #1148, opened 2021-03-25
Added Atari Lynx codes

goldug:
First of all, it's just one commit really, I just messed up the commits because I'm a Github noob...

But I was thinking it's about time Atari Lynx got added to the cheats database. I've started with three games so far, I'll probably add more soon. This is just to get it started.
I'm also planning to add other systems not in the database already, such as NeoGeo Pocket and Color.


RobLoach:
Thanks a lot! These are great. Happy to see more codes come in.

PR #1149, opened 2021-03-28
FFVII USA serial fix

bslenul:
The 3 discs had the same serial, should fix https://github.com/libretro/RetroArch/issues/12203 (at least for FFVII, the person mentioned Legend of Dragoon and RE2 but I can't reproduce with these games, they scan properly for me 🤷 )

Before (disc 1 being labeled as "Disc 3"):

image

After:

image


RobLoach:
Thanks a lot! Hopefully the next redump update, it'll get in too.

PR #1155, opened 2021-04-26
Removed duplicated serials from beta games for PS1 + deleted a duplicated uppercase .dat file for X360

bslenul:
`Microsoft - XBOX 360 (Digital).dat` being an outdated version of `Microsoft - Xbox 360 (Digital).dat` and it was causing issues because of the case on Windows, best to get rid of it.

I removed every duplicated serial for PS1 beta games that could conflict with retail games. Because as discussed on Discord, due to duplicated serials, a game might be scanned and added as a beta instead of a retail game.

I hope this is fine!


RobLoach:
Thanks! Made an issue over at https://github.com/RobLoach/libretro-dats/issues/43 to have them be skipped for EVERY system from no-intro and redump too. Thanks a lot!

PR #1156, opened 2021-05-02
Added accented characters to name and description

Jaffacakelover:
Altered spelling of Pokemon in all name and description fields, so that the word is written with the 'accent aigu' (as Pokémon). Rom names remain unchanged.


RobLoach:
Thanks for the fixes! Looks like you're adding a lot of changes? Was there a break in git?

Jaffacakelover:
Was there a break? I haven't a clue! I found and replaced the Pokemon titles (trying to leave the filenames un-accented), so there may be other titles with special characters that I didn't process.

RobLoach (changes requested):
+189,954 −35,861 .... Lots of additions that are likely not related.

PR #1157, opened 2021-05-03
GBA cheat codes cleanup!

davidgfnet:
This commit decrypts most of the encrypted CodeBreaker cheat codes into
their plain format. I manually fixed some games that had both plain and
encrypted codes, causing some of them not to work.

The advantage of having all codes in plain format is that it allows for
users to add their codes (appending the file). Encryption prevents this
since the cheat decoder tries to decrypt unconditionally.

Added a tiny python script that assists in decoding these cheatcodes. It
is not very smart so it might decrypt stuff that's already decrypted
depending on the order of the cheats in the file.

I manually eveyballed all the codes and ran some sanity checks on them
to the point I trust most of the changes here. Some codes were already
broken so I hopefully mend more things that I'm breaking with this
change.

Sorry for the whitespace cleanup! That's the script rewriting the files
cleanly


RobLoach (approved):
Nice work! Looking great :+1: ... While I am not versed on the core internals for cheats here, they seem like some good clean up. Thanks a lot.

PR #1173, opened 2021-06-25
Create Mega Man Battle Network (Code Breaker).cht

RumenBlack84:
cheat codes for megaman battle network 1
Seems to be missing from the repo, imported manually and they seem to work on my retroarch


RobLoach:
Thanks so much! :rocket:

PR #1174, opened 2021-07-09
Introduced new cheat files for Atari Lynx games

superKoder:
- Ninja Gaiden (1990)
- Rampage (1991)


cht/Atari - Lynx/Ninja Gaiden (1990).cht
```diff
@@ -0,0 +1,21 @@
+cheat0_address = "20244"
+cheat0_address_bit_position = "255"
+cheat0_big_endian = "false"
+cheat0_cheat_type = "1"
+cheat0_code = ""
+cheat0_desc = "my cheat"
```
RobLoach, line 6: Your cheat?

RobLoach (changes requested):
Thanks for ramping up the Lynx cheats!

cht/Atari - Lynx/Ninja Gaiden (1990).cht
superKoder: Good catch. Could have sworn I had fixed that. Will do!!

superKoder:
Here you go. Sorry for the delay, got busy at work.

PR #1175, opened 2021-07-10
Created dats for NEC PC-98 and Playstation Minis

Rstation4:
As for RPG Maker 3 Entries where added to the existing dat

### DIFF
--- a/dat/Atomiswave.dat
+++ b/dat/Atomiswave.dat
@@ -9,7 +9,7 @@ game (
 	name "Metal Slug 6"
 	description "Metal Slug 6"
 	genre "Action"
-	rom ( name mslug6.zip size 128344 crc 3582912721 131422139 md5 17b550ce0b58778d3ac304801be03768 sha1 637811d275a2716cd9abfe1d794bece650b65736 )
+	rom ( name mslug6.zip size 131422139 crc 3582912721 131422139 md5 17b550ce0b58778d3ac304801be03768 sha1 637811d275a2716cd9abfe1d794bece650b65736 )
 )
 
 game (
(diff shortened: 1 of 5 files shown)


Rstation4:
I already working on it again, needed to add more entries and fix info on a entry. Also not mentioned in the description but Atomiswave dat was created as well.

RobLoach:
Would this be run through Flycast? Mind adding the database entries to it's .info files in https://github.com/libretro/libretro-super/tree/master/dist/info ?

Rstation4:
> Would this be run through Flycast? Mind adding the database entries to it's .info files in https://github.com/libretro/libretro-super/tree/master/dist/info ?

Atomiswave and Sega - Naomi rom run with Flycast. Ive never added to the database entries so I'm not sure what to look for in there and what to edit.

dat/Atomiswave.dat
```diff
@@ -0,0 +1,76 @@
...
+	description "Atomiswave"
+	version 2021.7.9
+	author "libretro"
+)
+
+game (
+	name "Metal Slug 6"
+	description "Metal Slug 6"
+	genre "Action"
+	rom ( name mslug6.zip size 128344 crc 3582912721 131422139 md5 17b550ce0b58778d3ac304801be03768 sha1 637811d275a2716cd9abfe1d794bece650b65736 )
```
RobLoach, line 12: Are these .zip files torrentzipped? .zip files can have variable sizes.

Rstation4: The size is wrong on that one, it is a zip file at 131.4  MegaBytes. I will correct it with the following "131422139" I might as well review the ones I did again for Atomiswave as well so there's no mistakes.

PR #1183, opened 2021-08-09
Updated Diddy Kong Pilot

Jaffacakelover:
Resolved Game Boy Advance RDB. Diddy Kong Pilot (Proto) had 2 entries with the same name. Edited names and filenames. Checked files against CRC, MD5 and SHA1.


RobLoach:
Is this after an update to the .dat files?

Jaffacakelover:
Ah, I didn't find the GBA Dat before because there isn't one in `dat`... there's several in `metadat` though. Would it be just `metadat/no-intro/Nintendo - Game Boy Advance.dat` that needs updating to match?

RobLoach:
The `no-intro` ones are from https://datomatic.no-intro.org . We don't touch those ones as they're upstream. The changes are likely either in one of the other metadat folders. Apolgoies for the confusion.

PR #1186, opened 2021-09-06
Update Cannonball.dat

nsx0r:
updated because the previous .dat only listed 1 file (epr-10187.88) which is not enough to run the game.

epr-10381a.132 and epr-10381b.132 are included but not really needed, as they are the same as epr-10381.132 and the code can find them by their CRC value.
opr-10185.11 and opr-10186.47 are duplicates, this is normal and both are required.

sources: cannonball/roms/roms.txt & cannonball/src/main/roms.cpp


RobLoach:
Hmm, on another glance, I recognize this ends up having multiple roms.... libretro uses 1 game 1 rom format. We should just have one valid file in this list.

PR #1187, opened 2021-09-06
Update DATs

nsx0r:
checking zip archives makes no sense, so here's a dat that checks the actual roms for each games
added all 28 known/parents/available Atomiswave games (from present stable MAME source v0.235)
rebuild tested with clrmamepro, all games launch fine in latest retroarch/flycast core

### DIFF
--- a/dat/Atomiswave.dat
+++ b/dat/Atomiswave.dat
@@ -1,7 +1,7 @@
 clrmamepro (
 	name "Atomiswave"
 	description "Atomiswave"
-	version "2021.9.6"
+	version "2021.9.7"
 	author "libretro"
 )
 
@@ -12,9 +12,6 @@ game (
 	manufacturer "MOSS / Sammy"
 	genre "Sports"
 	rom ( name vm2001f01.u15 size 8388608 crc 2bb1be28 md5 384fcb0d63ec6f27e1a6d5b2d26d3999 sha1 fda7967d6c0341a608c52087ae3d461554760435 )
-	rom ( name vm2001f01.u2 size 8388608 crc 386070a1 md5 e988f6fa2272d3c14c2af59bed6ae481 sha1 bf46980ea822b4cfe67c622f0104bf793031f4ad )
-	rom ( name vm2001f01.u3 size 8388608 crc 4fb33380 md5 2612693f708207a2786fff3e23404e1b sha1 9070990515544e6e9a1d24b1e0597cdea926a4c9 )
-	rom ( name vm2001f01.u4 size 8388608 crc 7cb2e7c3 md5 d39ce04c27d211a35a0792e0526909e1 sha1 8b4e46cf19fbc1d613af75c52faebefb2776280b )
 )
 
 game (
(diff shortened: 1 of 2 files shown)


dat/Atomiswave.dat
```diff
@@ -1,82 +1,425 @@
...
+	manufacturer "Sega"
+	genre "Sports"
+	rom ( name vera.u1 size 16777216 crc cfef27e5 md5 5429059ef4d1cb379ff9b36e68303bf4 sha1 e0e27adc1b3635a310c50c6374d85572db608675 )
+	rom ( name vera.u14 size 16777216 crc 35df044f md5 8362c0fe4b5ab63f590b9624b3b4bd77 sha1 eeac6c4062f697205558846d6ac262cb5c1b10cf )
+	rom ( name vera.u15 size 16777216 crc e588afd1 md5 57baf84ab78f405292ee47ab128aab73 sha1 0ce3aeb2bcea66beaec2410d1df6857c4365aecf )
+	rom ( name vera.u16 size 16777216 crc 3590072d md5 ca7ab2e8aa0891796c024d59bf358909 sha1 3375a0334c35de1d7d8231d7cc27775451042f91 )
+	rom ( name vera.u17 size 16777216 crc d78389a4 md5 162e4a0ed8d4d178ecbb908929c62171 sha1 50babfe3d58929a26a69dd4a4120fd87f507a95e )
+	rom ( name vera.u2 size 16777216 crc 0a463c37 md5 440d8284959d040780231c29e0522c47 sha1 630ad98d2f80fd458729bd56e8d665a88263da28 )
+	rom ( name vera.u3 size 16777216 crc 8cbec9d7 md5 f2fdc3bcb542b761cfb093ae3137ead8 sha1 080f5edd817993946b1008ebe8ba489f818d3f99 )
+	rom ( name vera.u4 size 16777216 crc bd1f13aa md5 f40ce8ff7bac208063325ce0120d456d sha1 1ef9a7e684418baf8a61fef2610839fd72887d4c )
```
RobLoach, line 33: Note that libretro-database expects 1 rom per 1 game... So if you select a unique rom in each game, and target that, then the scanner will detect the rom itself, and run that.

RobLoach (changes requested):

dat/Atomiswave.dat
nsx0r: Maybe I got this wrong, are these files actually used by the program? I thought these DATs were for the end-users to use with clrmamepro and make sure they got all the correct files to run the games, just like DATs from no-intro, TOSEC or Redump. If so, what happens when the targeted file is in multiple archives? Does this mean each game MUST be identified with only 1 file, which then has to be unique across all games?

RobLoach: Yeah, RetroArch's scanner isn't too well equiped for that. From my testing, it ends up identifying the file with the first in the database, and then adds duplicates of the same for the rest.

nsx0r: Thank you for the clarification, I will trim the DAT, 1G1R style!
More to come :)

RobLoach:
This is incredible! Nicely done :+1: ... I know it's a lot of work to trim them down.

nsx0r (commented):
that should be good
am I supposed to do something else? i'm kinda new to github

PR #1206, opened 2021-09-14
Update Thomson - MOTO.dat

nsx0r:
Complete games set from latest TOSEC DATs for Thomson MO5 & MO6 & TO7 & TO8 & TO8D & TO9 & TO9+


nsx0r:
im tired

PR #1209, opened 2021-09-14
Add some Coleco cht files (conversion from mame)

toccata10:
As I wanted to play some coleco games on retroarch, I searched and converted some mame xml cheat files (source: http://cheat.retrogames.com/).
Just converting from hex to decimal was sufficient to create those working cht files.
Should not be too difficult to continue if anyone is interested by the entire coleco library.
The names are mame's names.


toccata10:
As I created some python scripts to automatically do the process, I'm closing to upload a better version

PR #1218, opened 2021-10-07
Fixed blank entries, added title

Jaffacakelover:
There were a number of "blank" entries in the database, some with nothing but a CRC hash: Tracked them down.
* Retained and filled out details for the entries that match No-Intro set and weren't already listed (mostly Betas).
* Deleted the entries that turned out to be duplicates: They had different hashes because they weren't from No-Intro (were other scene releases, not hacks/translation).

There are still a few more DB entries than files in the No-Intro releases, due to a few hacks and translation patches being included. Left these in the DB.

Also deleted leftover workfile from folder (Game Boy Advance (2021-08-02) diddy.rdb).

Finally, added entries for all of the Goodboy Galaxy Demos / Betas released so far.


RobLoach:
I believe the bad dumps are removed. Mind checking again??? THanks!

Jaffacakelover:
I thought I'd pushed an earlier commit before the 'filled out' lines were removed... no matter, I have backups. Out of 85 blank entries filled, 37 were Betas and 48 were scene releases. I've searched the latest commit for the scene hashes and they didn't show up, so consider that checked :-)

RobLoach:
Thanks! Seems fine to have these in the metadat folders.

PR #1227, opened 2021-11-05
Removed duplicate and non-working

IAMNOTRANA:
Some don't even work, many were duplicate and also added new cheat "Infinite Help"


RobLoach:
Thanks a lot for the clean up.

PR #1246, opened 2021-11-18
Update DOOM.dat

satorope:
- The MD5 of EXTRAS.WAD was missing, so I fixed the error when converting to RDB. I'm sorry.
- EXTRAS.WAD alone always loads DOOM2.WAD, so it was set with Ultimate DOOM (Unity) and Doom II NERVE (Unity).
- quakewiki.org does not version DOOM ( Unity ) WAD , only EXTRAS.WAD update differences , so I removed removed (v1.1).

### DIFF
--- a/dat/DOOM.dat
+++ b/dat/DOOM.dat
@@ -1,7 +1,7 @@
 clrmamepro (
 	name "Doom"
 	description "Doom"
-	version 2021.11.15
+	version 2021.11.18
 	author "Obiwantje, LodanZark, libretro"
 )
 
@@ -12,7 +12,7 @@ game (
 )
 
 game (
-	name "Adventures of Square, The - Episodes 1 & 2"
+	name "Adventures of Square, The - Episodes 1 & 2 (v2.1)"
 	description "Adventures of Square, The - Episodes 1 & 2 (v2.1)"
 	rom ( name SQUARE1.PK3 size 45202919 crc a2858888 md5 582CBE67EBD506DD55786552482099F8 sha1 A4BD92BDE933FCCC1761D749A2A0EE855F436C83 )
 )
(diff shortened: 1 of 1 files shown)


dat/DOOM.dat
```diff
@@ -309,6 +304,14 @@ game (
...
 	description "Doom II - No Rest for the Living (Unity)"
+	rom ( name EXTRAS.WAD size 64779 crc 8068229f md5 0c7caf25ad1584721ff5ecc38dec97a0 sha1 55a6a95c99d7fe7b9501d34b0afecd3252a629e1 )
+	rom ( name NERVE.WAD size 3821966 crc ab68447f md5 4214c47651b63ee2257b1c2490a518c9 sha1 321e951f6cc6e9d51bc16eac53e0d001f0f3f338 )
+)
+
+game (
+	name "Doom II - No Rest for the Living (Unity) (Update - September 3)"
+	description "Doom II - No Rest for the Living (Unity) (Update - September 3)"
+	rom ( name EXTRAS.WAD size 64587 crc 82bccddc md5 38e1bb6cdd1e4227eef13d22f6e35209 sha1 0f2a7139b640e19a08f384562f907cb0734b0512 )
 	rom ( name NERVE.WAD size 3821966 crc ab68447f md5 4214c47651b63ee2257b1c2490a518c9 sha1 321e951f6cc6e9d51bc16eac53e0d001f0f3f338 )
```
RobLoach, line 315: Note that RetroArch's RDBs take One-Rom-Per-Game. We'll need to target just one unique entry per game.

RobLoach (changes requested):

dat/DOOM.dat
satorope: In DOOM.DAT alone, Final Doom-Ejection and Strife-Veteran Edition include multiple WADs.
In the first place, EXTRAS.WAD is a PWAD that does not work by itself like VOICE.WAD, so it should be set with IWAD such as DOOM.DAT.
However, if IWAD is missing, Prboom adds it.
For example, Doom II - No Rest for the Living is PWAD, but at runtime it also loads DOOM2.WAD that is not specified in the RDB.
All you have to do is to make the appearance of RDB. Would you like to do the other things like CloneROM?

RobLoach: I'll merge and re-build the RDBs. Mind testing once they're in?

PR #1255, opened 2021-12-25
Initial "tooling" for DAT sorting

TobiX:
While trying to add some more games to the RPG Maker DAT file, I wondered if it would be a good idea to sort libretro's custom DAT files (to make it easier to find duplicates & try to avoid merge conflicts if multiple pull requests to the same file are pending)

This initial PR is just to see if this is something other people find valuable and to show how the initial sorting commit would look. If this is seen as something useful, I could put some more time into creating a GitHub Action which would always enforce proper order for a subset (or all) DAT files. (If the script is considered "too crude", I could also put some more time into a proper ClrMamePro DAT style checker...)

### DIFF
--- a/dat/HBMAME.dat
+++ b/dat/HBMAME.dat
@@ -2812,8 +2812,7 @@ game (
 	rom ( name decodemo.zip size 89178 crc a3a387d0 sha1 48ad6a75baa4774bea9c241f7f23403727b20ec5 )
 )
 game (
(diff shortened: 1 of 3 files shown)


RobLoach:
Seems like a great idea, thanks. You will find some of the dats were build from third-parties, so while I'd rather not modify those, it would be just fine to update the manually managed ones. Nicely done!

PR #1261, opened 2022-01-02
CHIP-8: Add CHIP-8 database

RobLoach:
Anyone know if there's a CHIP-8 database out there that we could pull from?


RobLoach:
https://github.com/JohnEarnest/chip8Archive/tree/master/roms

PR #1263, opened 2022-01-11
Update Sega - Game Gear.rdb

Jaffacakelover:
Filled out the 5 blank rows, matching hash values to Dat-O-Matic entries.
Then removed 4 of those rows as they were scene/bad releases.
Removed the old 'Frogger (Beta)' row and left the new one, because Dat-O-Matic says new row is a better dump of same rom.


RobLoach:
Would you mind submitting the .dat updates too??? :pray:

Jaffacakelover:
There are no blank lines in the DAT to fill out / remove, but I have updated the Frogger entry to the "better" one from the RDB.

RobLoach:
Thanks a lot. Removed the extra newlines for new builds over at https://github.com/RobLoach/libretro-dats/commit/1cd61e9c078e3f8fda5beaca14c90b8489b9ce23

PR #1266, opened 2022-01-17
Added some new cheat codes

picosonic:
Some cheat codes discovered through reverse engineering


RobLoach:
Thnaks for the contribution. Happy to see more work done on the cheats.

PR #1285, opened 2022-04-14
Added Amstrad GX4000 and Atari Jaguar Cheat Support

TheGent:
I have created cheat support especially for this using RetroArch and added them for everyone to enjoy.
I will also be creating for Amstrad CPC and any other core there after needing my attention.


RobLoach:
Tahnks a lot!

PR #1287, opened 2022-04-16
Added extensive iQue - Nintendo 64 Cheat support

TheGent:
I have created extensive iQue - Nintendo 64 cheats for the games that work. When Wave Race 64, Yohsi's Story etc are working i will create for those.


RobLoach:
These are great. Glad the iQue is getting some attention. Would it be okay if we named these as part of `Nintendo - Nintendo 64 (iQue)` or something? Just organizing them so that they appear closer to the n64 in the list.

TheGent:
Yea no problem

TheGent:
Ok i have renamed to Nintendo - Nintendo 64 (iQue) and managed to update the Mario Kart 64 (CHN) Support a bit more. Do i need to do another pull request?

RobLoach:
This is incredible. Big push forwards for cheats. Are you comfortable with these being merged?

TheGent:
Yea go for it [RobLoach](https://github.com/RobLoach)

PR #1296, opened 2022-06-04
Create Gran Turismo 2 USA (v1.2) (Duckstation).cht

IceLancerSR:
Full package of codes for this game (Duckstation) all credits to cookieplmonster! I just converted to retroarch format
All codes except 60fps are disabled by default. Enable per desire.
Its safe to enable them all. Tested.


RobLoach:
Thanks so much!

PR #1331, opened 2022-10-29
Cheats: Added SNES CrystalBeans: Infinite Lives

monoploid:
SNES CrystalBeans: Infinite Lives added


RobLoach:
Thanks for the fix!

PR #1354, opened 2023-05-01
Update metadat for no-intro, TOSEC, redump

zach-morris:
This needs review.  Rerunning and generating metadat files for using the latest no-intro, TOSEC, redump databases available, while also adding Sharp X1 (TOSEC), Atari 8-Bit (TOSEC/No-Intro), Commodore PET (TOSEC), Sharp X6800 (no-intro).

I found however in re-running that many old games are removed all together from the dat file, and indeed they're no longer in the source databases.  Unsure if the goal would be to match the source database, or only add new games that were not previously in the dat.


RobLoach:
Keeping in line with No-Intro/TOSEC is key, even if they remove old dumps. Thanks so much for the updates here, @zach-morris . Happy to have another able to run the updater! :rocket:

PR #1373, opened 2023-05-31
Commodore Amiga/CD32/CDTV WHDLoad+redump+TOSEC

### LINKED ISSUE #976: Retroplay amiga dats
Now that p-uae (supposedly) has support for the Retroplay set of whdload lha's (i'm assuming 'support' here means that the retroplay filenaming is [consistent enough to use their naming scheme](https://github.com/libretro/libretro-uae#games-that-need-a-specific-model)), using the retroplay dat sets for a whdload database (and maybe the hd database and unreleased too) would be nice.

They're in the root of the retroplay ftp.

Lack of standard naming to link the filename/checksums to game metadata is painful but could maybe be avoided by using the openretro game database https://openretro.org/ because it has most of the WHDLoads as game entries (many outdated though). 

Not sure it can be downloaded though.

That database also has more stuff that can be super useful for the core, though i don't know if cores can take up database info. For instance, [they recently added this.](https://openretro.org/2019/05/04/whdload-quit-key-can-be-registered)

Which is a major usability problem with whdload in emulators (where people just kill the emulator instead of shutting down, and then are surprised their saves are never written). If RA/the core could recognize this and send that keycode for a more or less graceful shutdown before the core is terminated, it would help a lot.

sonninnos:
The Amiga final rdb filesize is rather large thanks to TOSEC, but ZX Spectrum also is almost as large, so maybe it isn't that bad.. 

WHDLoad entry labels try to match No-Intro so that existing thumbnails work. Some cases might need refinining.

Closes  #976


RobLoach:
> The Amiga final rdb filesize is rather large thanks to TOSEC

I've found TOSEC includes EVERYTHING, including entries that noone would ever really want to run. We could consider just including `Commodore Amiga - Games - *` instead. Right now it'll include Demos and applications, which :shrug:

sonninnos:
Sure, let's do that.

RobLoach:
Another question, where is `dat/Commodore - Amiga.dat` from? Mind adding some documentation into the header of the .dat file so that others can update it later on? TThanks!

sonninnos:
It is Retroplay WHDLoad database. Header info added and TOSEC size halved.

i30817:
FYI, i recently improved [libretrofuzz](https://github.com/i30817/libretrofuzz) and was testing it for matching the libretro amiga thumbnails to the whdload retroplay, and had some better results after messing with the heuristics. 

It's still not perfect, and can't be because some games don't have covers, or you have to get rid of metadata to match ( '--before '_' --no--meta' 1rst run, and on a second run i add '--no-subtitle', and on a third run i try '--before '_' --no-meta --min 90 --delay-after 30' to catch the marginal matches and skip them if i want to). But it does get 90%+ there.

So that's a option if you don't want those 'clean names', or the database gets outdated. I wonder if the same matching mechanism could be used to automate keeping the mapping to date.

PR #1377, opened 2023-06-30
Create cheat for Spider-Man 3 (USA) GameShark.cht

Gtbita:
A cheat for Spider-Man 3 from Game boy advance


RobLoach:
Thanks a lot.

PR #1386, opened 2023-08-07
add BurgerTime (USA) cheat

snqwq:
I got bored a while back so I made some cheats for BurgerTime (USA version)
hope they can be useful or fun
I will maybe add more if I can get better at it

(if u have any tips on making cheats in RA please message me or comment)


RobLoach:
These are great! thanks so much. Happy to see more people involved in the cheats.

PR #1392, opened 2023-09-15
Improve vircon32 DAT and add its compiled RDB

vircon32:
This PR improves Vircon32 DAT with better game names, helping distinguish games from demos and tests programs, and sorting alphabetically. It also adds the compiled RDB for this DAT file.


RobLoach:
Thanks!

PR #1401, opened 2023-10-30
Add a few games/software to homebrew roms and fixing tabs

snqwq:
Added games from Inkbox and Morphcat as well as fixing some of the indentation being 4 spaces instead of a tab character

(if this isn't how you add ROMs so that the frontend can recognize them please tell me)


RobLoach:
Thanks so much, this is perfect. The RDBs will be built eventually to include these :+1:

PR #1410, opened 2023-12-08
Update and rename Spectravideo.dat to Spectravideo - SVI-318 - SVI-328.dat

WizzardSK:


RobLoach:
If we rename this, we will also have to update the playlist name that appears in RetroArch. Intentional?

WizzardSK:
I am not sure. I just wanted to correct the name to have "System - Name" format like other systems. Not sure where the playlist name is defined in RetroArch. If is it not a correct change, close that, please.

RobLoach:
I'm good with either one. Makes sense to me!

WizzardSK:
Ok, fixed that!

PR #1424, opened 2024-01-27
README.md update based on current situation

zoltanvb:
Updated as per current state of libretro-database and libretro-dats.
Most updates are straightforward, though NES is a bit unclear (but this is what should happen based on file titles), and for Thomson MO/TO, database has now Tosec instead of the bespoke converter so I added that.


zoltanvb:
+ I did not find any database that would currently use Trurip data, so mentions of that are removed.

RobLoach:
Thanks a lot for the updates here.

PR #1428, opened 2024-02-09
Add Enterprise - 128 Database

zoltanvb:
Custom conversion includes some name shortening and additions of genres from World of Spectrum database - EP128 shares most of the games with ZX Spectrum so it can be considered valid.


RobLoach:
Would it be out of line to have this merged in qith the ZX Spectrum database? Then we couldn't have to add the build-scripts and all.

zoltanvb:
Although the hardware is similar, EP128 is not binary compatible with ZX Spectrum, so that would cause confusion - files in this database can not be loaded by the ZX.

The huge number of "ZX Spectrum" entries in the file simply comes from the fact that it was relatively easy to do homebrew conversions, while [original SW list](https://wiki.enterpriseforever.com/index.php?title=Hivatalosan_megjelent_EP_j%C3%A1t%C3%A9kok_list%C3%A1ja) is not that long. I was considering restricting entries to have a "clean" DB, but that would really decrease the usefulness.

I can do PR for libretro-build-db.sh eventually (I would like to do another DB first, for Videoton TVC).

RobLoach:
Since this is from TOSEC, we could likely add it to the dats.json in https://github.com/robloach/libretro-dats . and have it automatically convert for us :+1:

zoltanvb:
It can be added there as well, but I put some effort into crafting the specific file in the PR, so it would be nice if this could stay as a basis in `dat/` , similar to other systems, and then the DB compilation could do the merge with the autoconverted Tosec data as usual.

If that is not wanted for some reason, then an automatic Tosec conversion is still better than nothing, I suppose.

RobLoach:
Sounds good with me. We'll need two things...
1. An entry in https://github.com/libretro/libretro-super/blob/master/libretro-build-database.sh to build the Enterprise - 128 database
2. A relevent .info file that has "Enterprise - 128" in the database field https://github.com/libretro/libretro-super/tree/master/dist/info

PR #1445, opened 2024-04-19
feat(database): add Halo: Combat Devolved

owenvoke:
I'm not 100% sure whether this should be versioned (with a new entry per version)? Or does it need to be a single entry, that the CRC, version, file name, etc. gets updated for?


RobLoach:
We usually index the most recent version, so this seems fine with me. Nicely done. Will be part of the database next time the RDB files are rebuilt.

PR #1450, opened 2024-05-08
feat(gb): add Capybara Quest

owenvoke:


RobLoach:
Thanks! I like the soundtrack.

PR #1467, opened 2024-07-12
Updated Golden Sun (USA, Europe).cht with working codes.

JamieDewhirst:
Whilst some of the old codes worked fine, there were many that didn't. (Infinite HP/PP, all the Djinn unlocks). I've now fixed the codes that were not working, added some additional codes for Max Stats per character and generally tidied up the ordering of the cheats to make more sense.


RobLoach:
Thanks a lot. The cheats could always use more attention

JamieDewhirst:
Cheers! Happy to help out when/where I can.

PR #1468, opened 2024-07-12
Create Final Fantasy I & II - Dawn of Souls (USA, Australia).cht

JamieDewhirst:
Added a .cht file for Final Fantasy I & II - Dawn of Souls. 

Have prefaced each cheat with [FFI] or [FFII] to indicate which of the two games the cheat is for.


RobLoach:
Seems like a good way to show the difference :+1:

PR #1493, opened 2024-11-02
CD32 database update

marrikhong:
Redump:
- coverdisk naming standarized (it was all over the place) - one pack of thumbs will cover both Redump and Tosec coverdisks
- Guardian, Diggers and Oscar, Brian the Lion - name modification to match the standard

TOSEC:
- Tosec is back,
- conflicts resolved,
- Games BIN/IMG/ISO/ + Coverdisks - matching thumbs packet is coming
- Games shows on the list as 'name (publisher)' for a clean look; extra data available as meta; tags added for easy TOSEC/REDUMP playlist creation

RDB:
- recompiled and tested


RobLoach (approved):
Looks great! Thanks. Would you mind adding the file definitions to the JSON file in this script? That way next time we run the script, it'll update the CD32 day for us. Thanks! https://github.com/robloach/libretro-dats

marrikhong:
Thanks! Updating the JSON now. Cheers!

PR #1494, opened 2024-11-03
Amstrad CPC update

zoltanvb:
Update to clean-cpc-db v1.2, with a slight customization: move first disk of multi-disk entries to last place. This ensures that scanned disks will be the ones that can be actually started.


RobLoach:
Do `game` entries with 2 or more `rom` entries work as intended?
```
game (
	name "Asterix Im Morgenland (Coktel Vision) (de)"
	description "Asterix Im Morgenland (Coktel Vision) (de)"
	rom ( name "Asterix Im Morgenland (1988)(Coktel Vision)(de)(Side B).dsk" size 195635 crc C1508616 md5 2819F0472BDADE793DB636ACF8AB7647 sha1 6317C29F8B7F7B9AC853A7B677DDB61610747C6A )
	rom ( name "Asterix Im Morgenland (1988)(Coktel Vision)(de)(Side A).dsk" size 198211 crc 31D37BA2 md5 76D6E4325EE129197B1E058EC2842C21 sha1 8DD69499A8BC654EF04D0B620629581234204539 )
)
```

zoltanvb:
In my experience, they do work, but only the last ROM entry will be recognized during scanning. So the reordering is done to ensure playlist will refer to the disk that can be loaded, same way as in previous update ( #1417 ).

RobLoach:
Thanks!

PR #1500, opened 2024-11-28
Remove Test Cartridges and BIOS entries

RobLoach:
- **Remove Test Cartridges**
- **Remove [BIOS] entries**

### DIFF
--- a/dat/Nintendo - Nintendo Entertainment System.dat
+++ b/dat/Nintendo - Nintendo Entertainment System.dat
@@ -5693,12 +5693,6 @@ game (
 )
 
 game (
-	name "Controller Test Program (Japan) (En) (Test Program)"
-	description "Controller Test Program (Japan) (En) (Test Program)"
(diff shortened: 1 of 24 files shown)


dat/Nintendo - Nintendo Entertainment System.dat
steven-sheehy, line 1: Why are there 31K additions and deletions to this file?

steven-sheehy (commented):
By my count there are 1274 entries in the RDBs without a name or rom_name. So I think we're missing a lot here. Think it would be better to keep all entries and move filtering to c_converter.

dat/Nintendo - Nintendo Entertainment System.dat
RobLoach: Possibly line endings?

RobLoach:
Definitely always good to add improvements to c_converter. Its source is handled through the RetroArch repository, so movement can be slower there than what we can accomplish here. For now, I'll merge this, and we can do some brainstorming about ways to improve c_converter. Thanks!

PR #1503, opened 2024-12-04
Fix numeric CRCs in DAT files

### LINKED ISSUE #1502: Non-hex CRCs
It seems the entries added in https://github.com/libretro/libretro-database/pull/1175/files added the CRC in its numeric format whereas every other entry uses hexadecimal encoding. You can tell since some entries are more than 4 bytes / 8 characters. 

I'm not completely sure but I think RetroArch will not support these numeric entries and if they're larger than 4 bytes they'll be [zeroed](https://github.com/libretro/RetroArch/blob/6475791e0702d509ca721d321f5bfcd10af68a56/database_info.c#L700C16-L700C35) out.

cc @RobLoach

steven-sheehy:
Fix #1502

### DIFF
--- a/dat/RPG Maker.dat
+++ b/dat/RPG Maker.dat
@@ -26,7 +26,7 @@ game (
 game (
 	name "Embric of Wulfhammer's Castle"
 	description "Embric of Wulfhammer's Castle"
-	rom ( name RPG_RT.ini size 2889959490 crc AC414842 md5 c7e1be4e3f5c55bf2030dae1939be88e sha1 b830be3e44eba9a4387b4df80cb8ec2ed632c136 )
+	rom ( name RPG_RT.ini crc AC414842 md5 c7e1be4e3f5c55bf2030dae1939be88e sha1 b830be3e44eba9a4387b4df80cb8ec2ed632c136 )
 )
 
 game (
--- a/dat/RPG Maker.dat
+++ b/dat/RPG Maker.dat
@@ -8,49 +8,49 @@ clrmamepro (
 game (
 	name ".flow"
 	description ".flow (ver.0.194)(English Patched)"
-	rom ( name RPG_RT.ini size 60192 crc 1204305322 md5 d89da02d37cc18fbe4224aeae13bfb76 sha1 df27d409e0b128dd6556e90677da67b016bcbbb9 )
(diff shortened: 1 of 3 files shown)


dat/RPG Maker.dat
```diff
@@ -8,49 +8,49 @@ clrmamepro (
...
 	name "Dessert Nightmare"
 	description "Desert Nightmare (English Translation)"
-	rom ( name RPG_RT.ini size 2009 crc 4282449154 md5 130853a12cbe7fa33b8259c72b13fb36 sha1 9f04c16c591351bee98e2053565f5bdbffdd5845 )
+	rom ( name RPG_RT.ini size 2009 crc FF40FD02 md5 130853a12cbe7fa33b8259c72b13fb36 sha1 9f04c16c591351bee98e2053565f5bdbffdd5845 )
 )
 
 game (
 	name "Embric of Wulfhammer's Castle"
 	description "Embric of Wulfhammer's Castle"
-	rom ( name RPG_RT.ini size 2889959490 crc 2889959490 md5 c7e1be4e3f5c55bf2030dae1939be88e sha1 b830be3e44eba9a4387b4df80cb8ec2ed632c136 )
```
steven-sheehy, line 29: This line looks suspect. Previously the size had the same value as the CRC. One of those has to be incorrect, but I don't know the upstream source is for this RPG Maker DAT to confirm.

RobLoach: Hah, yeah there's no way the size matches the CRC value. The size doesn't really matter for the dat file. It could likely just be removed.

dat/RPG Maker.dat
```diff
@@ -8,49 +8,49 @@ clrmamepro (
...
+	rom ( name RPG_RT.ini size 2889959490 crc AC414842 md5 c7e1be4e3f5c55bf2030dae1939be88e sha1 b830be3e44eba9a4387b4df80cb8ec2ed632c136 )
```
RobLoach, line 29:
``` suggestion
	rom ( name RPG_RT.ini crc AC414842 md5 c7e1be4e3f5c55bf2030dae1939be88e sha1 b830be3e44eba9a4387b4df80cb8ec2ed632c136 )
```

PR #1505, opened 2024-12-10
Remove BIOS and program entries

steven-sheehy:
Remove more BIOS and program entries from DATs. These entries were filtered out by logic in libretro-dats but still included in the final RDB due to the merge process. Confirmed they all had missing names and description fields in the RDB as a result.

There's still a lot more invalid entries for bad dumps and demos. I removed the ones in this PR manually but probably a tool that can filter and rewrite files would be needed to remove the bad or demo entries en masse.

Related to https://github.com/RobLoach/libretro-dats/issues/93

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 /libretro-super
+.DS_Store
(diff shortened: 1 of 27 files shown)


RobLoach:
Thanks so much. Those old metadat files could always use some updates :+1:

PR #1507, opened 2024-12-11
Fix Partners in Time (USA) Cheats

GABO14231:
The code for Max Beans was actually one for Max Blue Peppers, which had no code before. Besides correcting the name, an actual cheat for Max Beans was added.


RobLoach:
thanks

PR #1525, opened 2025-03-20
Rebuild DICE db to pick up an unbuilt dmy-file fix.

mittonk:
Must have failed to rebuild after a tweak, this binary now includes the dmy-file content hashes.


mittonk:
No, found zip file matching digging too deep as well, taking a step back.

PR #1524, opened 2025-03-13
DICE dat, generated by mittonk/libretro-database-dice.

mittonk:
Matches the known-good dumps in their usual packaging, and dummy files, on Linux, Mac, and Windows.
Aiming to get this in and iterate on platforms and edge cases in another branch.

PR #1526, opened 2025-03-20
Revert "Merge pull request #1524 from mittonk/dice_db"

mittonk:
This reverts commit 03d4d0c17ed2f4f305511378b414d6c0bb257d67, reversing changes made to d7e2e5d92cc9878131b74ea79733be969a1deb22.

The included rdb was stale and a little broken, and the included dat generates a worse one.  Time to get out of master and go back to the drawing board, sorry!


RobLoach:
Could we improve the existing one? What do you think is missing?

mittonk:
1. The rdb file matches all extant zips and generates good entries like "steeplechase.zip", good.
It includes some crc32=0000000 entries for pong.dmy et al that might cause false positives, though I haven't been able to trigger that.

So, not that bad.

2. The dat file tried to update to newer dmy launcher files, and matches them well as "pong.dmy", good.
However, it produces zip file matches like "steeplechase.zip#003773-A.C4" that doesn't work with the libretro core's file loading, so those entries in the playlist are broken

OK for now... as long as nobody generates a fresh rdb file from it.

Thoughts?

mittonk:
I think this is fixable, I'll work on a fix, discarding revert.

mittonk:
Took a breather and wrote a fix instead: https://github.com/libretro/libretro-database/pull/1527

PR #1530, opened 2025-04-08
[ScummVM] Added King's Quest 1

Jaffacakelover:
Added King's Quest I: Quest for the Crown (AGI/DOS) from GOG.com collection (King's Quest 1+2+3)


RobLoach:
Thanks, could you mind submitting the changes to the .dat files too?

RobLoach:
Perhaps a better way to handle this is through adding the games to this manual JSON file:
https://github.com/RobLoach/libretro-database-scummvm/blob/master/manual-games.json

Then next time the upstream Gruby AdventurePack gets updated, we could bring that in and aggregate it correctly. But for now, I'll merge this one in. Thanks.

RobLoach:
> I can add KQ1 and LSL3 to the JSON file you mention as well / instead?

That may be a good idea. We can also add a flag to catch those xx release year entries too.

Going to merge this in the mean time

PR #1533, opened 2025-04-18
AmigaJay 35-39 added

marrikhong:
AmigaJay 35-39 added; cd32 tosec/redump metadata restored; rdb compiled; tested


metadat/redump/Commodore - CD32.dat
```diff
@@ -1,1181 +1,1828 @@
...
 	homepage "http://github.com/robloach/libretro-dats"
 )
 
 game (
 	name "ATR - All Terrain Racing (Europe)"
 	description "ATR - All Terrain Racing (Europe)"
+	releaseyear 1995
+	publisher "Team 17"
 	region "Europe"
+	tags "CD32-REDUMP-Games"
```
RobLoach, line 14: Thanks for the additions here. Unfortunately the updates in redump and tosec have a risk of getting lost next time the dats are scraped. I'm not sure of a good solution for it either, other than moving the entries to another folder...

```
game (
	name "ATR - All Terrain Racing (Europe)"
	releaseyear 1995
	publisher "Team 17"
	tags "CD32-REDUMP-Games"
)
```

RobLoach (changes requested):
Not sure of the best solution here when we're making changes to the upstream.

metadat/redump/Commodore - CD32.dat
marrikhong: Hello Rob, 
CD32 TOSEC is dead, no updates after 2019 if I remember correctly. 
Up to you, I can move TOSEC + REDUMP to EXTRAS (libretro-database\dat) or just turn the automatic update off (for CD32 redump and tosec). Or just update the dats now and let the extra data to be overwritten with the next update. Looks like most people uses manual scan when adding roms anyway.

PR #1537, opened 2025-04-24
Update hashes for Dinothawr.zip

SpiralBrad:
Updates the hashes for Dinothawr.zip to match the current https://buildbot.libretro.com/assets/system/Dinothawr.zip

### DIFF
--- a/dat/System.dat
+++ b/dat/System.dat
@@ -91,7 +91,7 @@ game (
 	rom ( name JiffyDOS_1581.bin md5 20b6885c6dc2d42c38754a365b043d71 )
 
 	comment "Dinothawr"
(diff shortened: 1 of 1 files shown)


RobLoach:
Thanks so much

martinbark:
This does not work.  Zip files will not give consistent hashing even when the contents are exactly the same.  If you download https://buildbot.libretro.com/assets/system/Dinothawr.zip today it will not match the hash in this merge request despite the contents not changing.  You need to trrntzip Dinothawr.zip first to give consistent hashing. If you do this then the hash should be this

```
rom ( name Dinothawr.zip size 5775931 crc 684b7b1a md5 0126033d2e51e4a3962ea261fe2e6d79 sha1 9606280f65ae057a555fe23de54868e7ed7f3532 )
```

PR #1538, opened 2025-04-26
cheats: Add Super Nintendo 46億年物語 (繁)

jangfu:
46 億年物語 中文金手指


khanson679:
Should this not be ASCII like the other filenames? I noticed this because it caused a backup to fail.

Also, I don't see a game of this title in the database. (I also can't read the last character 繁.) The closest I can find is "46 Okunen Monogatari - Harukanaru Eden e (Japan)".

PR #1554, opened 2025-07-20
rdb2dat.py: convert RetroArch rdb to clrmamepro dat

anon-coward:
Example usage:

    mkdir dats
    ./rdb2dat.py ./*.rdb -o ./dats/

Code was generated with AI assistasnce and is based on:

 - RDBEd by Bernhard Schelling (https://github.com/schellingb/RDBEd)
 - RetroArch's libretro-db (https://github.com/libretro/RetroArch/tree/master/libretro-db)

### DIFF
--- a/scripts/rdb2dat.py
+++ b/scripts/rdb2dat.py
@@ -1,0 +1,285 @@
+#!/usr/bin/env python3
+# RDB to DAT Converter
+#
+# Based on:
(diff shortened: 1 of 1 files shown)


RobLoach (changes requested):
The source dat files are already in this repository, so I'm unsure of why we would want to re-convert from from RDB to .DAT?

PR #1555, opened 2025-07-29
Added English translation of Policenauts

Jaffacakelover:
Added English translation of Policenauts. ISO/WAV/CUE as created from following The Policenauts Translation Project instructions.

### DIFF
--- a/dat/Nintendo - Nintendo Entertainment System.dat
+++ b/dat/Nintendo - Nintendo Entertainment System.dat
@@ -26717,6 +26717,18 @@ game (
 )
 
 game (
+	name "Sweet Home (Japan) [T-En by Gaijin Productions v1.00]"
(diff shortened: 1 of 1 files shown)


RobLoach:
Mind providing the .dat file updates too? Thanks.

Jaffacakelover:
Done. I added one entry per disc instead of one entry per game, hope that's OK.

Jaffacakelover:
Also updated the NES RDB.
I plan to update the SNES RDB with `Bahamut Lagoon` English translations, but the DeJap roms give `bad checksum` so can't be run.

RobLoach:
I'm fine with adding these. Note that soft-patching works too. Saves from having to apply the patch directly. Just add it in the same folder:
https://docs.libretro.com/guides/softpatching/

PR #1561, opened 2025-08-06
Added games

Jaffacakelover:
Added Goodboy Galaxy (and demos), Celeste Classic 1 & 2, and Mother 3 translation patch (all versions).


Jaffacakelover:
There isn't a Dat file for GBA? This branch/pull should only cover GBA.

PR #1571, opened 2025-09-13
Fixing Pokemon Tags

ry10hu:
Edited with [schellingb/RDBEd](https://github.com/schellingb/RDBEd) and fixed all the Pokemon Franchise, Publisher, Genre, and Developer tags because they were bothering me.

Replaced the ones located in Retroarch/database/rdb with the edited ones and they work great.


RobLoach:
Mind sending the .dat file updates? Otherwise we'll lose your changes.

ry10hu:
Sorry, is there any way to decompile the RDB into .dat files?
I edited the database itself because it seemed a lot easier with RDBEd than manually editing the .dat files.
Or is there perhaps a tool to automate the process of editing the .dat? Getting each ROM hash manually sounds like a pain.

Thank you!

RobLoach:
I haven't used RDBed before, but I wouldn't be surprised if they've been thinking about adding an export feature.

> Also, is there a reason why there's 2 different folders with dat files? (libretro-database\dat and libretro-database\metadat*)

Metadat tends to be third-party import sources, while /dat is maintained locally.

ry10hu:
Cool, I'll get those generated and committed then!

PR #1574, opened 2025-09-30
Add Videoton TV-Computer Database

zoltanvb:
Manually converted + fixed from GameBase original.

The platform is supported by ep128emu-core.


zoltanvb:
Will do. I tried building manually, Database Manager can handle the resulting `.rdb`.

PR #1591, opened 2025-12-25
cheats: Add Sony - PlayStation Portable cheats

Emulayr:
Sony Playstation Portable cheats folder is almost empty compared to 'Sony - Playstation'. This PR is a convert of a cheats.db to .cht files supported by RetroArch. Database source, reddit post:
https://www.reddit.com/r/EmulationOniOS/comments/1dp836r/here_is_the_cheats_database_for_all_ppsspp_games/

Some of them tested on PPSSPP core, some not function (the same as in PSX cheats), but this could be a good start.


Emulayr:
This can be merged, had to remove/replace specials from filenames preventing repo cloning on Windows.

Thanks to @OctopusButtons for creating Playstation Portable folder in #1573 and to examples from Playstation folder on how to use emulator cheats, this set a ground for this PR.

There are some cheats without any effect whatsoever, would be good to investigate further why this is happening, but for now these files could be used as many codes are functioning.

PR #1594, opened 2026-01-01
Update Pokemon - FireRed Version (USA, Europe) (Rev 1).cht

williamsluke780-dev:


RobLoach:
I'm unclear on the enabled option. Shouldn't that be disabled everywhere?

PR #1674, opened 2026-06-12
Improve Touhou Fan Translations Metadata

Ar4m1s:
I saw that the version that is provided from [Moriya Shrine](https://moriyashrine.org/resources/categories/retro-era.6/) is different from the versions already in the "dat/NEC - PC-98.dat" file. I added these versions using this nushell script:
```nushell
#!/usr/bin/env nu
def main [] {
  fd -e hdi | lines | each {|e| sum $e}
}

def sum [filename: string] {
  $"rom \( name \"($filename | path basename)\" size (du $filename | $in.apparent | into int | last) crc (cksum -a crc32b $filename | split words | first | into int | format number | $in.upperhex | str substring 2.. ) md5 (md5sum $filename | split words | first) sha1 (sha1sum $filename | split words | first) \)"
}
```
I manually tested this using the explore view in RetroArch, and it worked fine. I just had to copy over the rdb file to the correct location.

I also added more metadata such as releaseyear, franchise, releasemonth to both the already existing translations and the ones I added.

I am unsure whether or not the titles are correctly set for three reasons:

1. The titles at Moriya Shrine are different as compared to the ones already in the file.
2. The titles already provided are inconsistent, but only in the description "Touhou Yumejikuu ~ the Phantasmagoria of Dim.Dream (English Translation)" instead of "Touhou Yumejikuu - the Phantasmagoria of Dim.Dream (English Translation)", why is it suddenly ~ and not - as in all other cases?
3. The titles are not the same on as on Wikipedia, "the" is added where Wikipedia lacks them.

There might be more metadata such as publisher and origin.

Lastly I want to mention that I am unsure whether or not I should add the resulting rdb file, espcially since when I saw that I got a different result for all the other rdb files as well (perhaps a version mismatch or similar).

### DIFF
--- a/dat/NEC - PC-98.dat
+++ b/dat/NEC - PC-98.dat
@@ -8,35 +8,100 @@ game (
 	name "Touhou Reiiden - Highly Responsive to Prayers"
 	description "Touhou Reiiden - Highly Responsive to Prayers (English Translation)"
 	genre "Break Out"
+	franchise "Touhou"
+	releaseyear "1997"
+	releasemonth "8"
 	rom ( name "TH01 - Highly Responsive to Prayers.hdi" size 5330944 crc 9883C55A md5 6750fc0224e0a4ba1d7e31a00dfae4f7 sha1 123fb01863a64c032a3d88790bde5bd27ffdd5b6 )
+)
+
+game (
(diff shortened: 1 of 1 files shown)


RobLoach:
Thanks!

PR #1677, opened 2026-06-14
Improve FBNeo dat generator script

Ar4m1s:
FBNeo_dat_gen.py has a few issues.

1. The file was encoded as dos, which was different from the rest of the scripts.
2. The script did not follow convention for cli args.
3. A function tried to use the parser variable which was not passed to the function leading to error.
4. When it generates the dat, it uses year instead of releaseyear.

This PR essentially just fixes these problems.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -2,3 +2,4 @@
 *.dat text eol=lf
 *.xml text eol=lf
 *.cht text eol=lf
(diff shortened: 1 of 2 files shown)


Ar4m1s:
The diff looks quite bad because of the encoding change.

PR #1678, opened 2026-06-14
Update FBNeo split dat file

Ar4m1s:
I reran the FBNeo_dat_gen.py script with a more updated version of the romset and xml dat file. I did find that there were a lot of roms missing that exist in the xml dat file that I did not have in my romset, but I don't think that is an issue, especially based on the lines added vs subtracted.


Ar4m1s:
I think I should look at some new sources.

Ar4m1s:
If I find the rest of the roms, I'll update it. This should at least be an improvement.

PR #1568, opened 2025-09-01
CHEATS for Project GaiaSeed and Taroumaru

OctopusButtons:
- Project GaiaSeed, a PS1 shooter
- Shinrei Jusatsushi Taroumaru, a Saturn action sidescroller that recently got a fan translation to English

PR #1684, opened 2026-06-23
cht: Add Infinite Lives (Player 2) for Shinrei Jusatsushi Taroumaru (Sega Saturn)

jancborchardt:
Follow-up to https://github.com/libretro/libretro-database/pull/1568 by @OctopusButtons, thanks a lot for that! :)
Now it also works if you want to play with 2 people.

(Before finding this existing cheat file, I looked for the reduction of a continues value. But lives makes more sense.)


jancborchardt:
(Force push to correct the copypaste mistake of having the new cheat prepended with `cheat0` as well instead of the needed `cheat1`.)

